PR #4, opened 2024-04-11
Updated README with build instructions and added hidden files to gitignore

liandra-giacomine:


Ewan-James-Donovan (dismissed):
LGTM! It may be worth putting something about starting up mongo before running sm2 start as the frontend will need mongo

ChrisCookOC (approved):
lgtm :)

PR #6, opened 2024-04-22
Tgp 449/create category guidance page

besscerule:


conf/messages.en
```diff
@@ -38,8 +38,12 @@ error.number = Please enter a valid number
 error.required = Please enter a value
 error.summary.title = There is a problem
 
-categoryGuidance.title = Setting up your profile
-categoryGuidance.h1 = Boo
+categoryGuidance.title = Categorisation - Trader Goods Profile
```
liandra-giacomine: I think this will cause the same issue as Nidhi found in the other ticket, where "- Trader Goods Profile" becomes duplicated in the title. I think we only need Categorisation here

besscerule: Changed

PR #15, opened 2024-04-24
TGP-680: Created session service to handle responses from session repository

liandra-giacomine:


app/models/UserAnswers.scala
```diff
@@ -82,10 +46,13 @@ object UserAnswers {
 
     (
       (__ \ "_id").write[String] and
-        (__ \ "data").write[JsObject] and
+        (__ \ "traderGoodsProfile").writeNullable[TraderGoodsProfile] and
         (__ \ "lastUpdated").write(MongoJavatimeFormats.instantFormat)
     )(unlift(UserAnswers.unapply))
   }
 
+//  val writes: OWrites[UserAnswers] = Json.writes[UserAnswers]
```
ChrisCookOC: Guess we can remove these now?

app/models/UserAnswers.scala
```diff
@@ -82,10 +46,13 @@ object UserAnswers {
...
+//  val reads: Reads[UserAnswers]    = Json.reads[UserAnswers]
+
```
besscerule: we need to remove this

app/services/SessionService.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class SessionService @Inject() (sessionRepository: SessionRepository)(implicit ec: ExecutionContext) {
+
+  def createUserAnswers(internalId: InternalId): EitherT[Future, SessionError, Unit] =
+    EitherT {
+      val emptyUserAnswers = UserAnswers(internalId.value)
+      sessionRepository
```
besscerule, line 32: could we use updateUserAnswers in here? or is that a faff?

app/controllers/NiphlQuestionController.scala
```diff
@@ -16,43 +16,59 @@
...
-            //TODO change redirect to page after niphl number in the journey
-            routes.DummyController.onPageLoad.url
-          }
-          Redirect(url)
+        formWithErrors => Future.successful(BadRequest(view(formWithErrors))),
+        hasNiphlAnswer => {
+          val updatedTgpModelObject = request.userAnswers.traderGoodsProfile.copy(hasNiphl = Some(hasNiphlAnswer))
+          val updatedUserAnswers    = request.userAnswers.copy(traderGoodsProfile = updatedTgpModelObject)
+
+          sessionService
```
ChrisCookOC, line 61: Not worth spending more than a few seconds thinking about, but going through I keep thinking can we extract this to some sort of "saveAnswers" routine. 
But I'm totally happy with how this is anyway so feel free to ignore

it/test/helpers/ItTestBase.scala
```diff
@@ -64,10 +65,16 @@ trait ItTestBase extends PlaySpec with GuiceOneServerPerSuite {
     Enrolment("HMRC-CUS-ORG").withIdentifier("fake-identifier", Gen.alphaNumStr.sample.get)
   private val authResult              = Some("internalId") and Enrolments(Set(ourEnrolment))
 
-  def authorisedUser: OngoingStubbing[Future[Option[String] ~ Enrolments]] =
-    when(authConnector.authorise(any, eqTo(authFetch))(any, any)).thenReturn(
+  def authorisedUserWithAnswers: OngoingStubbing[Future[Option[String] ~ Enrolments]] = {
```
ChrisCookOC: We can just change this to not return the mock I think. We only returned the mock originally because it was the last thing that happened. 

Looks like we can return Unit and end the routine with () to return "Unit". Not sure if that is better/clearer than returning the mock

test/controllers/NirmsQuestionControllerSpec.scala
```diff
@@ -34,11 +38,17 @@ class NirmsQuestionControllerSpec extends SpecBase {
...
+    formProvider,
+    sessionRequest,
+    sessionService
   )
 
   "NirmsQuestion Controller" - {
 
+    nirmsQuestionController.onPageLoad(
+      fakeRequest
+    ) // TODO: MOCK Properly, Prepare some answers using sessionRequest and SessionService proper mocks.
```
ChrisCookOC: is this todo for a later ticket?

test/services/SessionServiceSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    "createUserAnswers should create answers if sessionRepository does not fail" in {
+      when(
+        mockSessionRepository.set(any())
+      ).thenReturn(
+        Future.successful(true)
+      )
+
+      val result: EitherT[Future, SessionError, Unit] = sessionService.createUserAnswers(id)
+      result.value.futureValue match {
```
ChrisCookOC: Might be nicer to rewrite this as 
result.value.futureValue mustBe Right() 
rather than having to map each value?

Although if the concern is more being explicit in the test, could wrap it in a clue

withClue("Session service should not fail to create answers"){
    result.value.futureValue mustBe Right() 
}

it/test/helpers/ItTestBase.scala
besscerule: So this might be a problem for later on - but how do you manually manipulate the useranswers object in this - currently it generates a full useranswers in the test but we might not always want that

test/controllers/NirmsQuestionControllerSpec.scala
besscerule: presumably the 'write tests properly ticket' will cover all of this

besscerule (changes requested):

test/controllers/NirmsQuestionControllerSpec.scala
Ewan-James-Donovan: it was a leftover comment, i can remove

test/services/SessionServiceSpec.scala
Ewan-James-Donovan: Cool! I'll try that

Ewan-James-Donovan: Yep that's better, I'll implement that

Ewan-James-Donovan: Done!

it/test/helpers/ItTestBase.scala
Ewan-James-Donovan: I agree, Chris - I'll do that and say done when done. And Beth, this is just here since the tests will fail if the user has no answers (will redirect to the profile setup page, as per other logic implemented). The answers are empty and I think we can iterate as and when need comes along. If there we other cases it's just a matter of putting "authorisedUserWithSomeParticularAnswers" -> obviously not with that as the variable name though... perhaps with a simpler refactor, too.

Ewan-James-Donovan: Done!

app/controllers/NiphlQuestionController.scala
Ewan-James-Donovan: Probably, but for now this will suffice, feel free to make a ticket on it, since that would be a nice to have thing :)

besscerule (approved):

ChrisCookOC (approved):

PR #16, opened 2024-04-24
TGP-431 Nirms Question Page

ChrisCookOC:


README.md
```diff
@@ -21,15 +21,17 @@ Use service manager to start up supporting services: `sm2 --start TGP_FE`
...
+Run this to check integration tests: `sbt it/test`
 
 Run this to check code coverage: `sbt clean coverage test`
 
 Run this to get code coverage report: `sbt coverageReport`
 
 ### Formatting
 
-Run this to format scala files : `sbt scalafmt`
+Run this to format scala files : `scalafmt`
```
liandra-giacomine: should be `sbt scalafmtCheckAll`

app/views/NirmsQuestionView.scala.html
```diff
@@ -0,0 +1,52 @@
...
+    @paragraph(Html(messages("nirmsQuestion.p1")))
+    @paragraph(Html(messages("nirmsQuestion.p2", link(messages("nirmsQuestion.p2.linkText"),"https://www.gov.uk/government/publications/retail-movement-scheme-how-the-scheme-will-work/retail-movement-scheme-how-the-scheme-will-work#product-eligibility"))))
+
+    @heading2(messages("nirmsQuestion.h2"))
+
+    @formHelper(action = NirmsQuestionController.onSubmit) {
+        @govukRadios(
+          RadiosViewModel.yesNo(
+                field = form("value"),
+                legend = Legend()
```
liandra-giacomine: The Legend shouldn't be empty, we can use `nirmsQuestion.h1` for this

ChrisCookOC: That would change the design so it no longer looks like the prototype though? 

with the legend
image

vs prototype
image

README.md
ChrisCookOC: Done

app/views/NirmsQuestionView.scala.html
liandra-giacomine: I think there’s a way to hide it, .asHiddenSomething
I recall it being something we had to update in a prior project for accessibility reasons

ChrisCookOC: ok yeah while looking up how to do it I found the accessibility repo saying how its a common accessibility error! 

Updated :)

mycp98 (dismissed):
all looks good to me 🙂

liandra-giacomine (approved):

liandra-giacomine (changes requested):
will re-approve once ukims is updated to redirect to this page

it/test/controllers/NirmsQuestionControllerISpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import helpers.ItTestBase
+import play.api.http.Status.{BAD_REQUEST, OK, SEE_OTHER}
+import play.api.libs.ws.{WSClient, WSRequest}
+import play.api.test.Helpers.{await, defaultAwaitTimeout}
+
+class NirmsQuestionControllerISpec extends ItTestBase {
+
+  lazy val client: WSClient = app.injector.instanceOf[WSClient]
+
+  private val url = s"http://localhost:$port${routes.NirmsQuestionController.onPageLoad.url}"
```
besscerule, line 28: given this is a test it shouldn't be relying on stuff in the controller because that could be wrong! i think it should be hard coded

besscerule (changes requested):

it/test/controllers/NirmsQuestionControllerISpec.scala
ChrisCookOC: It is relying on the routes not the controller, so I'm not too worried about it because we aren't trying to test the routes have been setup correctly here? 

But I have no strong opinions, we switched to this way after another code review comment :D

app/views/NirmsQuestionView.scala.html
```diff
@@ -0,0 +1,52 @@
...
+
+@(form: Form[Boolean])(implicit request: Request[_], messages: Messages)
+
+@layout(
+    pageTitle    = titleNoForm(messages("nirmsQuestion.title")),
+    showBackLink = true
+) {
+    @heading1(messages("nirmsQuestion.h1"))
+    @paragraph(Html(messages("nirmsQuestion.p1")))
+    @paragraph(Html(messages("nirmsQuestion.p2", link(messages("nirmsQuestion.p2.linkText"),"https://www.gov.uk/government/publications/retail-movement-scheme-how-the-scheme-will-work/retail-movement-scheme-how-the-scheme-will-work#product-eligibility"))))
```
besscerule: could we split this over 2 lines for readability

it/test/controllers/NirmsQuestionControllerISpec.scala
besscerule: leave it i'm just being picky :p

app/views/NirmsQuestionView.scala.html
ChrisCookOC: Done

besscerule (dismissed):

Ewan-James-Donovan (dismissed):
LGTM

it/test/controllers/NirmsQuestionControllerISpec.scala
liandra-giacomine: @besscerule  The url needs to be correct for the test to call the right endpoint, not hardcoding it and instead setting it to the routes url means we won't have to update this test if we change the path

liandra-giacomine (dismissed):

besscerule (dismissed):

yohan-oc (approved):
Looks good to me

Ewan-James-Donovan (approved):
LGTM :)

PR #21, opened 2024-05-01
TGP 435 Create CYA Page

besscerule:
Ewan: Refactored the helper and allowed for language changing. Uses messages file site.change


app/controllers/KeepAliveController.scala
```diff
@@ -27,15 +27,15 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  getData: SessionRequestAction,
   sessionRepository: SessionRepository
 )(implicit ec: ExecutionContext)
     extends FrontendBaseController {
 
   def keepAlive: Action[AnyContent] = (authorise andThen getData).async { implicit request =>
-    request.userAnswers
-      .map { answers =>
-        sessionRepository.keepAlive(answers.id).map(_ => Ok)
+    request.userAnswers.traderGoodsProfile.ukimsNumber
```
liandra-giacomine: why are we passing the ukims number here?

liandra-giacomine: Oh I see, I think previously userAnswers was optional in the request whereas it's not optional only need to call the keep alive function here and it seems that the way it's handling errors is by ignoring them..returning an Ok.

It does raise the question though, shouldn't we be using the session service  here, but that would require more work we we'd need to handle errors appropriately (at the moment exceptions are not being caught)

app/controllers/helpers/CheckYourAnswersHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+      ),
+      createOptionalYesNoSummaryListRow(
+        routes.NirmsQuestionController.onPageLoad.url,
+        "NIRMS registered",
+        traderGoodsProfile.hasNirms
+      ),
+      createOptionalSummaryListRow(
+        routes.NirmsNumberController.onPageLoad.url,
+        "NIRMS number",
+        traderGoodsProfile.nirmsNumber match {
```
liandra-giacomine: A simpler way to do this: 
`traderGoodsProfile.ukimsNumber.map(_.value)`

If it's None, it will remain as None, if there's a Some of Ukims then we can extract the value in the Some and convert it to a Some[String] in the map

app/controllers/helpers/CheckYourAnswersHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      Some(createSummaryListRow(href, key, value.get))
+    }
+
+  private def createOptionalYesNoSummaryListRow(
+    href: String,
+    key: String,
+    value: Option[Boolean]
+  ): Option[SummaryListRow] =
+    if (value.isEmpty) {
```
liandra-giacomine: Could use map here aswell, as it does the same thing as the if statement
    ```
value.map { booleanValue =>
      createSummaryListRow(href, key, if (booleanValue) "Yes" else "No")
    }
```

File: app/views/NiphlQuestionView.scala.html [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+    link: components.Link
+)
+
+@(form: Form[Boolean])(implicit request: Request[_], messages: Messages)
+
+@layout(
+    pageTitle    = titleNoForm(messages("niphlQuestion.title")),
+    showBackLink = true,
+// TODO session data check for nirms
+    backLinkUrl = if ("we have a nirms") controllers.routes.NirmsNumberController.onPageLoad.url else controllers.routes.NirmsQuestionController.onPageLoad.url
```
liandra-giacomine: we might get to a point where we have complex logic for the back link, would be good to extract the functionality of getting the url to it's own component but it's not something to worry about right now

liandra-giacomine (commented):
Looking good! As discussed, we'll need to add query params to the endpoints for the Mode and split the logic accordingly in the controllers. It may also be useful to implement cleasing functions (in the companion object?) for when we need to set None to values etc :)

app/controllers/KeepAliveController.scala
besscerule: This is from Ewans branch I think :)

app/controllers/helpers/CheckYourAnswersHelper.scala
besscerule: done

app/views/NiphlQuestionView.scala.html
besscerule: removed for now because they in makes more sense for the checkyour answers stuff back and fourth

app/controllers/helpers/CheckYourAnswersHelper.scala
besscerule: done

app/controllers/KeepAliveController.scala
Ewan-James-Donovan: my bad if theres an issue here, I had a lot to juggle

Ewan-James-Donovan:
Overall seems good to me :) Will approve if we're not sorting keep alive stuff here

test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -17,30 +17,92 @@
...
+    sessionRequest,
+    mockCheckYourAnswersHelper
+  )
+
+  "CheckYourAnswersController" - {
 
     "must return OK and the correct view for a GET" in {
 
-      val application = applicationBuilder(userAnswers = emptyUserAnswers).build()
+      val summaryList = List(
```
Ewan-James-Donovan, line 49: We could shorten this using the helper and creating answers... but this is fine as is and it could be argued that this keeps helper and controller test seperate.

besscerule: i like to separate things - we could have separate test-utils file called test-gubbins or the like where we can store the huge variables

will keep it as it is unless you really want it to change

conf/app.routes
```diff
@@ -5,32 +5,35 @@
...
 
-GET         /ukims-number                                controllers.UkimsNumberController.onPageLoad
-
-POST        /ukims-number                                controllers.UkimsNumberController.onSubmit
-
-GET         /nirms-question                              controllers.NirmsQuestionController.onPageLoad
+GET         /ukims-number                                controllers.UkimsNumberController.onPageLoad(mode: Mode = NormalMode)
+POST        /ukims-number                                controllers.UkimsNumberController.onSubmit(mode: Mode = NormalMode)
+GET         /ukims-number/check                          controllers.UkimsNumberController.onPageLoad(mode: Mode = CheckMode)
+POST        /ukims-number/check                          controllers.UkimsNumberController.onSubmit(mode: Mode = CheckMode)
```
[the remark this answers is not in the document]
Ewan-James-Donovan: It seems to work in a similar case for me

test/controllers/CheckYourAnswersControllerSpec.scala
Ewan-James-Donovan: Sounds good! No need to change for now. If it works and is well tested, it works. - I tend to leave personal preference outside of the "need to have" realm.

app/controllers/helpers/CheckYourAnswersHelper.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package controllers.helpers
+
+import controllers.routes
+import models.{CheckMode, TraderGoodsProfile}
+import uk.gov.hmrc.govukfrontend.views.Aliases._
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.Key
+
+class CheckYourAnswersHelper {
```
Ewan-James-Donovan: I simplified the implementation here to handle the optionals cleaner and instead map through a sequence of summary data to make the summary rows. No big deal - I just wanted to - it was fine how it was.

It can be extended nicely in the cases to accommodate for other answer types.

Ewan-James-Donovan (dismissed):

Ewan-James-Donovan:
needs a merge with main

app/controllers/NiphlNumberController.scala
```diff
@@ -56,15 +56,16 @@ class NiphlNumberController @Inject() (
       .fold(
         formWithErrors => Future.successful(BadRequest(view(formWithErrors))),
         niphlNumber => {
-          val updatedMaintainProfileAnswers =
-            request.userAnswers.maintainProfileAnswers.copy(niphlNumber = Some(NiphlNumber(niphlNumber)))
-          val updatedUserAnswers            = request.userAnswers.copy(maintainProfileAnswers = updatedMaintainProfileAnswers)
+          val updatedTgpModelObject =
+            request.userAnswers.maintainProfileAnswers
+              .copy(hasNiphl = Some(true), niphlNumber = Some(NiphlNumber(niphlNumber)))
```
liandra-giacomine: I don't think we should override hasNiphl here, it is possible that they will enter this value but never fill out the question page in which case we should fail with an error. We need a validation check when they submit in CYA to ensure there are no empty values for ukims and questions/answers for numbers they stated they have

app/controllers/NirmsNumberController.scala
```diff
@@ -40,30 +40,35 @@ class NirmsNumberController @Inject() (
...
       .fold(
-        formWithErrors => Future.successful(BadRequest(view(formWithErrors))),
+        formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode))),
         nirmsNumber => {
-          val updatedMaintainProfileAnswers =
-            request.userAnswers.maintainProfileAnswers.copy(nirmsNumber = Some(NirmsNumber(nirmsNumber)))
-          val updatedUserAnswers            = request.userAnswers.copy(maintainProfileAnswers = updatedMaintainProfileAnswers)
+          val updatedTgpModelObject =
+            request.userAnswers.maintainProfileAnswers
+              .copy(hasNirms = Some(true), nirmsNumber = Some(NirmsNumber(nirmsNumber)))
```
liandra-giacomine: same here

app/controllers/helpers/CheckYourAnswersHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+            actions = Some(Actions(items = Seq(ActionItem(href, HtmlContent(changeText)))))
+          )
+        )
+      case _                     => None
+    }
+  }
+  def createSummaryList(
+    traderGoodsProfile: MaintainProfileAnswers
+  )(implicit messages: Messages): List[SummaryListRow] = {
+    val summaryData: Seq[(String, String, Option[Any])] = Seq(
```
liandra-giacomine: You can just set this to a list to begin with so you don't have to set it .toList later

test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -17,37 +17,93 @@
...
+      val result = checkYourAnswersController.onPageLoad()(fakeRequest)
+
       status(result) mustEqual OK
 
-      contentAsString(result) mustEqual checkYourAnswersView(expectedList)(fakeRequest, messages).toString
+      contentAsString(result) mustEqual checkYourAnswersView(list)(fakeRequest, stubMessages()).toString()
 
     }
 
+    "must redirect on Submit" in {
```
liandra-giacomine, line 99: can we add a test for the unhappy path/s

app/controllers/helpers/CheckYourAnswersHelper.scala
Ewan-James-Donovan: true!

app/controllers/helpers/CheckYourAnswersHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+        )
+      case booleanValue: Boolean =>
+        Some(
+          SummaryListRow(
+            key = Key(HtmlContent(key)),
+            value = Value(HtmlContent(if (booleanValue) "Yes" else "No")),
+            actions = Some(Actions(items = Seq(ActionItem(href, HtmlContent(changeText)))))
+          )
+        )
+      case _                     => None
```
liandra-giacomine: This is an interesting case - if it doesn't match any of the types here this will not write the row in the table....

besscerule: should this cause an error?

test/controllers/CheckYourAnswersControllerSpec.scala
besscerule: I'm not sure there is an unhappy path atm because it just directs to the dummy controller regardless on submit

app/controllers/helpers/CheckYourAnswersHelper.scala
besscerule: done

app/controllers/NiphlNumberController.scala
besscerule: can we chat about this

app/controllers/helpers/CheckYourAnswersHelper.scala
besscerule: do we want to change this or keep it?

app/controllers/NiphlNumberController.scala
besscerule: chatted - removed

app/controllers/NirmsNumberController.scala
besscerule: removed

app/controllers/helpers/CheckYourAnswersHelper.scala
Ewan-James-Donovan: I thought it would be best to allow the developer to add new types, there would be a match error if they started developing say a number input (not a number string)... We could remove this case, raise an error in this case, or keep it as is... I'd be inclined to keep it so the match cases are exhaustive, an error would be fine also.

Ewan-James-Donovan (dismissed):
Me likey

app/views/UkimsNumberView.scala.html
```diff
@@ -27,19 +28,18 @@
...
 )
 
-@(form: Form[_])(implicit request: Request[_], messages: Messages)
+@(form: Form[_], mode: Mode)(implicit request: Request[_], messages: Messages)
 
 @layout(
     pageTitle    = titleNoForm(messages("ukimsNumber.title")),
-    showBackLink = true,
-    backLinkUrl = controllers.routes.ProfileSetupController.onPageLoad.url
+    showBackLink = true
```
yohan-oc, line 35: This will not work now since we delete the JS back button behaviour in a previous ticket https://github.com/hmrc/trader-goods-profiles-frontend/commit/d30688f1dfd44fcc3d4a5c8114ff9170537d15e2

@liandra-giacomine Shall I fix this in my ticket https://jira.tools.tax.service.gov.uk/browse/TGP-801 ?

besscerule: What's the plan for this? Am happy to add the change?

app/controllers/helpers/CheckYourAnswersHelper.scala
liandra-giacomine: An error would be fine except we know which types we allow so it wouldn't make sense to provide invalid types/overcomplicate the logic unnecessarily imo. We realised that all of these get converted to string so instead of doing the conversion through type matching, we can just expect string and make the conversion to string prior to this :)

app/views/UkimsNumberView.scala.html
liandra-giacomine: Back links aren't being tested yet but I am pro having stuff that works - whether we want to add the JS button functionality back or add backLinkUrl = controllers.routes.ProfileSetupController.onPageLoad.url back in I don't mind so long as it isn't causing a failure, as this will be worked on in a future ticket

test/controllers/helpers/CheckYourAnswersHelperSpec.scala
```diff
@@ -0,0 +1,301 @@
...
+
+      "must return empty summary list when blank trader goods profile" in {
+        val maintainProfileAnswers: MaintainProfileAnswers = MaintainProfileAnswers()
+        val summaryList                                    = checkYourAnswersHelper.createSummaryList(maintainProfileAnswers)(fakeMessages)
+        summaryList mustBe Seq.empty
+      }
+
+      "must return full summary list when trader goods profile has all fields" in {
+        val expected                                       = List(
+          SummaryListRow(
```
liandra-giacomine: Could you extract the creation of these rows to a function so it's easier to read?

app/controllers/NirmsNumberController.scala
```diff
@@ -40,30 +40,35 @@ class NirmsNumberController @Inject() (
...
+              .copy(nirmsNumber = Some(NirmsNumber(nirmsNumber)))
           val updatedUserAnswers            = request.userAnswers.copy(maintainProfileAnswers = updatedMaintainProfileAnswers)
 
           sessionService
             .updateUserAnswers(updatedUserAnswers)
             .fold(
               sessionError => Redirect(routes.JourneyRecoveryController.onPageLoad().url),
-              success => Redirect(routes.NiphlQuestionController.onPageLoad.url)
+              success =>
+                mode match {
```
liandra-giacomine: Might be nice to put this in its own private function in the controller
`private def navigate(mode: Mode) = mode match....`

app/views/UkimsNumberView.scala.html
yohan-oc: @besscerule you can see the change here https://github.com/hmrc/trader-goods-profiles-frontend/commit/d30688f1dfd44fcc3d4a5c8114ff9170537d15e2

app.js 

`document.addEventListener('DOMContentLoaded', function(event) {

   // handle back click
   var backLink = document.querySelector('.govuk-back-link');
   if (backLink !== null) {
     backLink.addEventListener('click', function(e){
       e.preventDefault();
       e.stopPropagation();
       window.history.back();
     });
   }
 });`

Layout.scala.html

`@(pageTitle: String, showBackLink: Boolean = true, timeout: Boolean = true, showSignOut: Boolean = true)(contentBlock: Html)(implicit request: Request[_], messages: Messages)`

`@if(showBackLink) {
         @govukBackLink(BackLinkViewModel(href = "#").withCssClass("js-visible"))
     }`

app/controllers/NirmsNumberController.scala
besscerule: done and added this to thje other controllers too - i agree it looks nicer

app/views/UkimsNumberView.scala.html
besscerule: fixed

test/controllers/helpers/CheckYourAnswersHelperSpec.scala
besscerule: is this what you mean? - check changes

yohan-oc (approved):

test/controllers/helpers/CheckYourAnswersHelperSpec.scala
liandra-giacomine: I was thinking more like extracting this bit:

```
def createRow(key: String, value: String, actionItem: String, content: String)
SummaryListRow(
       Key(HtmlContent(key)),
       Value(HtmlContent(value)),
       "",
       Some(
         Actions(
           "",
           List(ActionItem(actionItem, HtmlContent(content), None, "", Map()))
         )
       )
     )
```

Then create a sequence like `Seq(createRow(a,b,c), createRow(d,e,f)...)` but what you did makes it easy to see what is happening so happy with that! Approving PR

liandra-giacomine (approved):

PR #24, opened 2024-05-01
TGP-371: Create Commodity code page

vinoocs:


test/controllers/CommodityCodeControllerSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+
+      val result = commodityCodeController.onSubmit(fakeRequest)
+
+      status(result) mustEqual BAD_REQUEST
+
+      contentAsString(result) mustEqual commodityCodeView(formWithErrors)(fakeRequest, stubMessages()).toString
+
+    }
+
+    "must send bad request on Submit when user enters invalid commodity code" in {
```
liandra-giacomine: it's hard to know whether this is testing the correct thing because both conditions for this and the below are the same (bad request and view) - could add check for specific error code or something

liandra-giacomine (commented):
Looking great :) QA will have to be updated that the identifier is changing in this ticket so they update their tests

test/controllers/CommodityCodeControllerSpec.scala
vinoocs: It's just another test case, one for checking if it's invalid format, other for checking if it does not fit 6,8 or 10 digits. I have added the error code asset though :)

test/forms/CommodityCodeFormProviderSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+  val validCommodityCodeGenerator: Gen[String] = {
+    val validLengthsGen = Gen.oneOf(6, 8, 10)
+    validLengthsGen.flatMap(length => Gen.listOfN(length, Gen.numChar).map(_.mkString))
+  }
+
+  val invalidCommodityCodeGenerator: Gen[String] = {
+    val invalidLengthGen = Gen.choose(1, 12).suchThat(len => len != 6 && len != 8 && len != 10)
+    val invalidCharsGen  = Gen.alphaNumStr.suchThat(_.exists(!_.isDigit))
+
+    Gen.oneOf(
```
ChrisCookOC, line 39: Nice stuff with the Gens, you managed to make more sense of them then I did! Will hopefully be able to use some of these combining them patterns going forward :)

liandra-giacomine (changes requested):
Can you pull the latest from main and add session handling?

test/controllers/CommodityCodeControllerSpec.scala
Ewan-James-Donovan: I think this does cover the cases, perhaps be more clear in the name of the test

vinoocs:
> Can you pull the latest from main and add session handling?

Done it!

yohan-oc (approved):
Looks good to me

Ewan-James-Donovan (approved):
Good stuff :)

liandra-giacomine (approved):

PR #27, opened 2024-05-07
TGP 831: Message Spec and IT tests

vinoocs:
- Message spec has been added
- Validated messages in Integration tests
- Improved test coverage


test/utils/MessagesSpec.scala
```diff
@@ -0,0 +1,190 @@
...
+    val missingKeys = expectedKeys.diff(messages.keySet)
+    withClue(s"In $label, the following keys are missing: ${missingKeys.mkString(", ")}") {
+      missingKeys.isEmpty mustBe true
+    }
+  }
+
+  private lazy val englishMessages: Map[String, String] = getExpectedMessages("en") -- commonMessageKeys
+
+  private def getExpectedMessages(languageCode: String) =
+    messagesAPI.messages.getOrElse(languageCode, throw new Exception(s"Missing messages for $languageCode"))
```
liandra-giacomine: Instead of throwing an exception here you can return an Option, where if it's empty you fail the test:
Use Map[String, Option[String]], then value match { 
case Some(value) => ...
case None => fail("Missing messages for...")

vinoocs: Thanks, updated now :)

liandra-giacomine (dismissed):

liandra-giacomine (approved):

PR #28, opened 2024-05-07
TGP-878 increase code coverage to 90

besscerule:


README.md
```diff
@@ -27,9 +27,7 @@ Run this to check integration tests: `sbt it/test`
 
 Run this to check accessibility tests: `sbt clean a11y:test`
 
-Run this to check code coverage: `sbt clean coverage test`
-
-Run this to get code coverage report: `sbt coverageReport`
+Run this to check all code coverage and get report: `sbt clean coverage test it/test a11y:test coverageReport`
```
besscerule, line 30: This includes ALL the tests

We probably want to put a11y:test into pipeline - what do we think?

build.sbt
```diff
@@ -70,4 +70,7 @@ lazy val it =
     .enablePlugins(PlayScala)
     .dependsOn(microservice % "test->test")
 
-addCommandAlias("runAllChecks", ";clean;compile;scalafmtCheckAll;coverage;test;it/test;scalastyle;coverageReport")
+addCommandAlias(
+  "runAllChecks",
+  ";clean;compile;scalafmtCheckAll;coverage;test;it/test;a11y:test;scalastyle;coverageReport"
```
besscerule, line 75: it doesn't like a11y:test, wants me to do a11y/test but that breaks it :(

test/base/SpecBase.scala
```diff
@@ -60,7 +60,23 @@ trait SpecBase
...
 
-  val sessionRequest = new FakeSessionRequestAction(emptyUserAnswers)
+  val emptySessionRequest = new FakeSessionRequestAction(emptyUserAnswers)
+
+  private val categorisationAnswers =
+    CategorisationAnswers(Some(CommodityCode("anything")), Some(CountryOfOrigin("GB")))
+
+  private val maintainProfileAnswers =
+    MaintainProfileAnswers(
+      Some(UkimsNumber("anything")),
```
besscerule, line 70: should these be valid numbers? do we care for testing purposes?

test/base/SpecBase.scala
```diff
@@ -76,6 +92,19 @@ trait SpecBase
     Future.successful(Right(()))
   )
 
+  val badSessionService: SessionService = mock[SessionService]
+
+  when(badSessionService.readUserAnswers(any[InternalId])) thenReturn
```
liandra-giacomine: why are we doing this here? :)

besscerule: Because it's used in a lot of different specs - is there a better way?

build.sbt
besscerule: Also - i don't think it includes a11y in the test coverage report

besscerule: actually scrub that - it is

README.md
ChrisCookOC: Yeah in pipelines sounds sensible to me

test/base/SpecBase.scala
ChrisCookOC: probably but I've not been doing it either. Since we are not testing the values of the field directly it probably doesn't matter right now?

test/base/SpecBase.scala
ChrisCookOC: I think my instinct would be to not put mocks in the base, because its hard to tell in the individual test where the mocking is happening. Could be middle ground, e.g. a function that has all the mocks in and you can call that from each spec?

besscerule: this is done for the sessionService mock above it though - do we removed that too?

ChrisCookOC: I guess ideally we'd just have the one sessionService mock and not different ones for each case

And in the tests where we need it to return nothing, set the mock to return nothing, in the tests where we need badsessiondata you set it to do that etc.

besscerule: I like this idea - much better

test/base/SpecBase.scala
Ewan-James-Donovan: I share Chris's opinion.

test/controllers/NiphlNumberControllerSpec.scala
```diff
@@ -56,6 +60,44 @@ class NiphlNumberControllerSpec extends SpecBase with NiphlNumberGenerator {
 
     }
 
+    "must return OK and the full view for a GET" in {
+
+      val fullNiphlNumberController = new NiphlNumberController(
```
Ewan-James-Donovan, line 65: If we can put this full controller next to the other empty one at the top, that'd be nice, but it'm not fussed if you're happy.

Ewan-James-Donovan (dismissed):
I think this is good, I understand clearly what you're doing and if it works, it works! Please resolve conflicts and I'll reapprove

test/controllers/NiphlNumberControllerSpec.scala
besscerule: i was going to day because it is only used in one test keep it within that scope?

Ewan-James-Donovan: Fine by me!

Ewan-James-Donovan (approved):
Reapproved

ChrisCookOC (approved):

PR #31, opened 2024-05-08
TGP-457 - Create homepage

ChrisCookOC:


a11y/FrontendAccessibilitySpec.scala
```diff
@@ -1,4 +1,4 @@
-import models.{CheckMode, Mode, NormalMode}
+import models.{Mode, NormalMode}
```
Ewan-James-Donovan: Do we want accessibility for check mode, too? Maybe not a question for this ticket.

app/controllers/CheckYourAnswersController.scala
```diff
@@ -43,9 +46,14 @@ class CheckYourAnswersController @Inject() (
...
-  // TODO replace dummy route and post session data
-  def onSubmit: Action[AnyContent] = authorise { implicit request =>
-    Redirect(routes.DummyController.onPageLoad.url)
+  def onSubmit: Action[AnyContent] = (authorise andThen getData).async { implicit request =>
+    routerService
+      .setUpProfile(request.eori, request.userAnswers.maintainProfileAnswers)
+      .fold(
+        error => Redirect(routes.DummyController.onPageLoad.url), //TODO redirect to actual error page
+        success => Redirect(routes.HomepageController.onPageLoad.url)
+      )(catsStdInstancesForFuture(defaultExecutionContext))
```
Ewan-James-Donovan: Is the catsStdInstancesForFuture used for implicit excution context? I'm used to seeing `class XyzClass @Inject() ( <some injects>) (implicit ec ExecutionContext)`. I'm not averse to this cats way if we know it works properly.

it/test/base/WireMockServerSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.github.tomakehurst.wiremock.WireMockServer
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration.options
+
+trait WireMockServerSpec {
+
+  implicit lazy val wireMock: WireMockServer = new WireMockServer(options().dynamicPort())
+
+  val wireMockHost: String   = "localhost"
+  lazy val wireMockPort: Int = wireMock.port()
+}
```
Ewan-James-Donovan, line 28: Love the use of wiremock

Ewan-James-Donovan (dismissed):
Looks good to me, I'll reapprove after conflicts are fixed, no major comments,

a11y/FrontendAccessibilitySpec.scala
ChrisCookOC: Was wondering that as I removed the import! I think right now the pages are so similar in each mode its not a big concern, but maybe something for down the line

app/controllers/CheckYourAnswersController.scala
ChrisCookOC: nah so it wants an implicit Functor[Future], which is the catsStdInstancesForFuture, which then itself wanted an execution context.

Have moved it up out of the method call and it looks a bit nicer I think. Let me know if that is clearer

a11y/FrontendAccessibilitySpec.scala
Ewan-James-Donovan: Cool!

app/controllers/CheckYourAnswersController.scala
Ewan-James-Donovan: That's clearer :)

Ewan-James-Donovan (approved):
Lgtm

besscerule (approved):

besscerule:
Looks ace :)

PR #32, opened 2024-05-08
TGP 895 cya bug fix heading

besscerule:


conf/messages.en
```diff
@@ -130,9 +130,20 @@ countryOfOrigin.inputHeading = Pick a country
...
+checkYourAnswers.title = Check your answers
+checkYourAnswers.h1 = Check your answers
+checkYourAnswers.h2 = UKIMS, NIRMS and NIPHL
+checkYourAnswers.ukimsNumber = UKIMS number
+checkYourAnswers.nirmsNumber = NIRMS number
+checkYourAnswers.nirmsRegistered = NIRMS registered
+checkYourAnswers.niphlNumber = NIPHL number
+checkYourAnswers.niphlRegistered = NIPHL registered
+
 checkYourAnswers.h2 = UKIMS, NIRMS and NIPHL
```
vinoocs, line 143: Duplicates may be

vinoocs: Duplicates may be, line 143,144 & 145

test/controllers/helpers/CheckYourAnswersHelperSpec.scala
```diff
@@ -32,6 +32,11 @@ class CheckYourAnswersHelperSpec extends SpecBase with SummaryListFluency {
   when(fakeMessages("site.change")) thenReturn "Change"
```
ChrisCookOC, line 32: Seems unnecessary to mock all the messages here? We can just use the stubbed messages in the SpecBase and check the result match the values like "site.change"

mycp98 (approved):

test/controllers/helpers/CheckYourAnswersHelperSpec.scala
besscerule: i agree! -changed! :)

besscerule: actually - having checked your way makes the tests fail - if you can fix it better in your ticket feel free - but I'll pop this one in for now :)

PR #35, opened 2024-05-09
TGP-888 Validate answers for maintain profile

besscerule:


yohan-oc (dismissed):

ChrisCookOC (approved):

liandra-giacomine (commented):
I think this PR can be closed as this was implemented in Nick's PR and we decided not to use an action for it as it is not a component that will be reused so it can just be implemented in the controller

PR #56, opened 2024-05-13
Cya validation

besscerule:


app/controllers/CheckYourAnswersController.scala
```diff
@@ -44,35 +49,49 @@ class CheckYourAnswersController @Inject()(
...
+            NiphlNumberSummary.row(request.userAnswers)
+          ).flatten
+        )
+        Ok(view(list))
+      }.getOrElse {
+        logErrorsAndContinue(maybeErrors)
+      }
   }
 
+  def incorrectHasNirms(userAnswers: UserAnswers): Boolean = userAnswers.get(HasNirmsPage).contains(false)  && userAnswers.get(NirmsNumberPage).isDefined
```
liandra-giacomine: this isn't being used anywhere

test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -38,37 +40,98 @@ class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency wi
 
     "for a GET" - {
 
-      "must return OK and the correct view" in {
+      "must return OK and the correct view with valid non-optional data" in {
```
liandra-giacomine, line 43: It took me a bit to understand this description and the one below.
Maybe update to something more specific like with mandatory fields present
and the one below to with optional fields present? Not sure exactly how to best word it but it's a little confusing to me right now :D

liandra-giacomine: nitpick: maybe more intuitive to state with mandatory data rather than non-optional :D

app/controllers/CheckYourAnswersController.scala
besscerule: This is for discussion with Nick later today - the special cases

test/controllers/CheckYourAnswersControllerSpec.scala
besscerule: agreed

liandra-giacomine (dismissed):

app/models/TraderProfile.scala
```diff
@@ -31,26 +31,24 @@ object TraderProfile {
...
-  private def getNiphl(answers: UserAnswers): IorNec[ValidationError, Option[String]] =
-    answers.getIor(HasNiphlPage).flatMap {
-      case true  =>
-        answers.getIor(NiphlNumberPage).map(Some(_))
-      case false =>
-        if (answers.isDefined(NiphlNumberPage)) Ior.Left(NonEmptyChain.one(UnexpectedPage(NiphlNumberPage))) else Ior.Right(None)
+  private def getNiphl(answers: UserAnswers): EitherNec[ValidationError, Option[String]] =
+    answers.getEither(HasNiphlPage) match {
+      case Right(true)  => answers.getEither(NiphlNumberPage).map(Some(_))
+      case Right(false)  => if (answers.isDefined(NiphlNumberPage)) Left(NonEmptyChain.one(UnexpectedPage(NiphlNumberPage))) else Right(None)
```
liandra-giacomine: we can probably extract this bit to user answers and modify the function name to be more intuitive, maybe something like:

```
def unexpectedValueDefined(page: Gettable[_]):  = (answers.isDefined(page)) Left(NonEmptyChain.one(UnexpectedPage(page))) else Right(None)


def getPageValue[A](page: Gettable[A])(implicit rds: Reads[A]): EitherNec[ValidationError, A] =
   get(page).map(Right(_)).getOrElse(Left(NonEmptyChain.one(PageMissing(page))))
```

test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -38,37 +40,98 @@ class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency wi
...
 
         running(application) {
           val request = FakeRequest(GET, routes.CheckYourAnswersController.onPageLoad.url)
 
           val result = route(application, request).value
 
           val view = application.injector.instanceOf[CheckYourAnswersView]
-          val list = SummaryListViewModel(Seq.empty)
+          val list = SummaryListViewModel(
+            rows = Seq(
```
liandra-giacomine: the generation of rows seems to be duplicated, let's extract it to a function:

```
def getRows(userAnswers) = Seq(UkimsNumberSummary.row(userAnswers)(messages(application)),
               HasNirmsSummary.row(userAnswers)(messages(application)),
               NirmsNumberSummary.row(userAnswers)(messages(application)),
               HasNiphlSummary.row(userAnswers)(messages(application)),
               NiphlNumberSummary.row(userAnswers)(messages(application))
             ).flatten

val list = getRows(userAnswers)
```

test/models/TraderProfileSpec.scala
```diff
@@ -82,10 +83,11 @@ class TraderProfileSpec extends AnyFreeSpec with Matchers with TryValues with Op
             .set(HasNirmsPage, true).success.value
             .set(HasNiphlPage, false).success.value
 
-        val (errors, data) = TraderProfile.build(answers).pad
+        val result = TraderProfile.build(answers)
 
-        data must not be defined
-        errors.value.toChain.toList must contain only PageMissing(NirmsNumberPage)
+        inside(result) { case Left(errors) =>
```
liandra-giacomine, line 88: looks cool :D What does inside(result) do?

liandra-giacomine: guessing it fails the test if it doesn't match on the partial function?

liandra-giacomine: Yep, it does fail if it doesn't match on the partial, nice!! love it

Ewan-James-Donovan (approved):
:)

PR #69, opened 2024-05-15
TGP-457 - Stop empty EORI being used

ChrisCookOC:


yohan-oc (dismissed):

vinoocs (dismissed):

app/controllers/actions/IdentifierAction.scala
```diff
@@ -57,9 +57,9 @@ class AuthenticatedIdentifierAction @Inject() (
           authorisedEnrolments
             .getEnrolment(config.tgpEnrolmentIdentifier.key)
             .flatMap(_.getIdentifier(config.tgpEnrolmentIdentifier.identifier)) match {
-            case Some(enrolment) =>
-              block(IdentifierRequest(request, internalId, enrolment.value))
-            case None            => throw InsufficientEnrolments("Unable to retrieve Enrolment")
+            case Some(enrolment) if enrolment.value.isBlank => throw InsufficientEnrolments("EORI must be defined")
```
liandra-giacomine: Some(enrolment) if enrolment.value.isBlank and None could go in the same block as they both throw the same exception. The thing that distinguishes them here is the message but that isn't being used anywhere

ChrisCookOC: Updated :)

liandra-giacomine (approved):

vinoocs (approved):

PR #71, opened 2024-05-16
TGP-368 Create goods description page

Ewan-James-Donovan:
Scaffold offers no TextArea input, so I used HmrcCharacterCount, which supports form errors and character count warnings. I think this is standard.

Also, do we have a standard way of doing page titles (the title that appears in the tab title)

image

### DIFF
--- a/app/views/GoodsDescriptionView.scala.html
+++ b/app/views/GoodsDescriptionView.scala.html
@@ -37,6 +37,10 @@
 
     @formHelper(action = routes.GoodsDescriptionController.onSubmit(mode)) {
 
+        <h1 class="govuk-heading-xl">@messages("goodsDescription.h1")</h1>
+        <p class="govuk-body">@messages("goodsDescription.p1")</p>
+        <p class="govuk-body">@messages("goodsDescription.p2")</p>
+
         @if(form.errors.nonEmpty) {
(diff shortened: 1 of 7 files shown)


app/views/GoodsDescriptionView.scala.html
```diff
@@ -30,16 +32,24 @@
...
+        <p class="govuk-body">@messages("goodsDescription.p1")</p>
+        <p class="govuk-body">@messages("goodsDescription.p2")</p>
+
         @if(form.errors.nonEmpty) {
             @govukErrorSummary(ErrorSummaryViewModel(form))
         }
 
-        @govukInput(
-            InputViewModel(
+        @textArea(
```
vinoocs, line 43: There's <textarea class="govuk-textarea" , not sure if that suffices the need though ):

liandra-giacomine (approved):
Looks good Ewan, I think it's confusing to have some navigation being done in some tickets and not in others but since you've started it here that's alright. Can you add a comment to the navigation ticket saying this has been done but testing is still needed?

app/views/GoodsDescriptionView.scala.html
liandra-giacomine: If we can use a HMRC pre defined model, this should be okay :)

vinoocs (approved):

PR #78, opened 2024-05-17
TGP-848 Commodity code result page

yohan-oc:


app/connectors/OttConnector.scala
```diff
@@ -0,0 +1,60 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package connectors
+
+import config.Service
```
ChrisCookOC, line 19: There are some unused imports in here

app/controllers/HasCorrectGoodsController.scala
```diff
@@ -44,23 +45,29 @@ class HasCorrectGoodsController @Inject() (
     extends FrontendBaseController
     with I18nSupport {
 
-  private val form = formProvider()
+  private val form             = formProvider()
+  private val defaultCommodity = Commodity("Default Code", "Default Description")
```
ChrisCookOC: is there a chance we display these to the user?

app/controllers/CommodityCodeController.scala
```diff
@@ -49,23 +53,31 @@ class CommodityCodeController @Inject() (
...
+              case UpstreamErrorResponse(_, NOT_FOUND, _, _) =>
+                val formWithApiErrors =
+                  form.copy(errors = Seq(elems = FormError("value", getMessage("commodityCode.error.invalid"))))
+                BadRequest(view(formWithApiErrors, mode))
+            }
         )
-  }
+    }
+
+  private def getMessage(key: String)(implicit messages: Messages): String = messages(key)
```
besscerule, line 82: why is this a separate function?

besscerule (changes requested):

app/controllers/CommodityCodeController.scala
Ewan-James-Donovan: Implicit messages were behaving weirdly and it became easier for us to do it this way, if it's a deal breaker then it probably is fairly simple to change.

app/connectors/OttConnector.scala
yohan-oc: This is fixed

app/controllers/CommodityCodeController.scala
besscerule: that's fine

besscerule (dismissed):

app/controllers/HasCorrectGoodsController.scala
yohan-oc: No user will not see.

Why I have it first place is I need pass the commodity object to fe

`val commodity = request.userAnswers.get(CommodityCodePage).getOrElse(defaultCommodity)

    Ok(view(preparedForm, mode, commodity))`
I can make it `Commodity("", "")`

Instead of `.getOrElse(defaultCommodity)` is there a better way to do this? What do you think?

Actually `.getOrElse(defaultCommodity)` is a invalid user scenario.

mycp98 (dismissed):

app/controllers/HasCorrectGoodsController.scala
Ewan-James-Donovan: Can we make the view accept an optional commodity Instead and then not do the getOrElse, call me if you want to talk about it.

Ewan-James-Donovan (changes requested):

app/controllers/HasCorrectGoodsController.scala
Ewan-James-Donovan: We made it optional :)

Ewan-James-Donovan (approved):

PR #82, opened 2024-05-21
TGP-1015: Validate CYA for create record

besscerule:


app/controllers/CyaCreateRecordController.scala
```diff
@@ -0,0 +1,66 @@
...
+import logging.Logging
+import models.{GoodsRecord, ValidationError}
+import play.api.i18n.{I18nSupport, MessagesApi}
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
+import uk.gov.hmrc.play.bootstrap.binders.RedirectUrl
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
+import viewmodels.govuk.summarylist._
+import views.html.CyaCreateRecordView
+
+class CyaCreateRecordController @Inject() (
```
liandra-giacomine, line 32: Is it possible to put each journey into a package e.g.
profile package would contain all profile related files and the cya page for that
Similarly here we could have a createRecord pakage  - I wonder if we do that whether we can rename these to CheckYourAnswers as it'll be under a different package - don't worry about it if it's too much effort

test/controllers/CyaCreateRecordControllerSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+          )
+
+          status(result) mustEqual OK
+          contentAsString(result) mustEqual view(list)(request, messages(application)).toString
+        }
+      }
+
+      "must return OK and the correct view with all data (including optional)" in {
+
+        val userAnswers = UserAnswers(userAnswersId)
```
liandra-giacomine: Generating userAnswers can be extracted to a function, if we need it in a lot of specs it might be good to create a Generator class and link it to the SpecBase, so we can do something like:


`def userAnswers(traderReference, commodityCode, countryOfOrigin...)`

test/navigation/NavigatorSpec.scala
```diff
@@ -226,28 +231,28 @@ class NavigatorSpec extends SpecBase {
...
 
             val answers =
               UserAnswers("id").set(HasNiphlPage, true).success.value.set(NiphlNumberPage, "1234").success.value
-            navigator.nextPage(HasNiphlPage, CheckMode, answers) mustBe routes.CheckYourAnswersController.onPageLoad
+            navigator.nextPage(HasNiphlPage, CheckMode, answers) mustBe routes.CyaNirmsNiphlController.onPageLoad
           }
         }
 
-        "to CheckYourAnswersPage when answer is No" in {
+        "to CyaNirmsNiphl when answer is No" in {
```
liandra-giacomine: CyaNirmsNiphl - is this what we've named the profile set up CYA controller?

besscerule: Yes :)

app/controllers/CyaCreateRecordController.scala
besscerule: i've been doing the moving packages thing and it is quite a huge job with lots of file changes, at the very least i don't think it wants to be part of this ticket. if we want to restructure i think we should do this as a separate ticket and check with the other devs about what structure would suit us - also i think that since it goes against the scaffold pattern we probably might not even want to do it

Ewan-James-Donovan (dismissed):
I like, would reapprove if there were improvements like the ones I think you're doing now :)

test/controllers/CyaCreateRecordControllerSpec.scala
besscerule: I've thought a lot about this and having been faffing around with it on the branch. I think it makes it clearer to just set the useranswers in the test as each test needs slightly different useranswers. There is benefit in having empty user answers and maybe mandatory and mandatory + optional, but the others are specific test cases, so it's probably better just to have them in the test

test/navigation/NavigatorSpec.scala
besscerule: Changed name

test/controllers/CyaCreateRecordControllerSpec.scala
besscerule: I've changed them in both CYA specs, take a look

app/controllers/CyaCreateRecordController.scala
besscerule: Agreed to not do this and leave it

app/navigation/Navigator.scala
```diff
@@ -32,12 +32,13 @@ class Navigator @Inject() () {
...
     case NirmsNumberPage         => _ => routes.HasNiphlController.onPageLoad(NormalMode)
     case HasNiphlPage            => navigateFromHasNiphl
-    case NiphlNumberPage         => _ => routes.CheckYourAnswersController.onPageLoad
+    case NiphlNumberPage         => _ => routes.CyaCreateProfileController.onPageLoad
     case TraderReferencePage     => _ => routes.HasGoodsDescriptionController.onPageLoad(NormalMode)
     case HasGoodsDescriptionPage => navigateFromHasGoodsDescription
     case GoodsDescriptionPage    => _ => routes.CountryOfOriginController.onPageLoad(NormalMode)
     case CountryOfOriginPage     => _ => routes.CommodityCodeController.onPageLoad(NormalMode)
     case CommodityCodePage       => _ => routes.HasCorrectGoodsController.onPageLoad(NormalMode)
+    case HasCorrectGoodsPage     => _ => routes.CyaCreateRecordController.onPageLoad
```
besscerule: Ticket for navigation needs to include this

app/views/CyaCreateProfileView.scala.html
```diff
@@ -23,20 +23,20 @@
...
+    <h1 class="govuk-heading-l">@messages("cyaCreateRecord.h1")</h1>
+    <h2 class="govuk-heading-m">@messages("cyaCreateRecord.h2")</h2>
     @govukSummaryList(list)
     <div class="govuk-warning-text">
         <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
         <strong class="govuk-warning-text__text">
-            <span class="govuk-visually-hidden">@messages("checkYourAnswers.warning")</span>
-            @messages("checkYourAnswers.warningText")
+            <span class="govuk-visually-hidden">@messages("cyaCreateRecord.warning")</span>
+            @messages("cyaCreateRecord.warningText")
```
yohan-oc: All `cyaCreateRecord` should replace with `cyaCreateProfile`

app/views/CyaCreateProfileView.scala.html
```diff
@@ -23,20 +23,20 @@
...
         </strong>
     </div>
 
-    @formHelper(action = routes.CheckYourAnswersController.onSubmit) {
+    @formHelper(action = routes.CyaCreateRecordController.onSubmit) {
```
yohan-oc: Controller should be `CyaCreateProfileController`

app/views/CyaCreateProfileView.scala.html
besscerule: fixed

app/views/CyaCreateProfileView.scala.html
besscerule: fixed

liandra-giacomine (approved):

PR #100, opened 2024-05-30
TGP-1087: Saving commodity code in query and page

vinoocs:
In this ticket, the separation of commodity code that comes from API has been been achieved.

### DIFF
--- a/app/models/GoodsRecord.scala
+++ b/app/models/GoodsRecord.scala
@@ -40,8 +40,8 @@ object GoodsRecord {
       getCommodityCode(answers),
       answers.getPageValue(CountryOfOriginPage),
       getGoodsDescription(answers)
-    ).parMapN((eori, traderReference, commodity, countryOfOrigin, goodsDescription) =>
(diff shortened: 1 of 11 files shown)


app/models/GoodsRecord.scala
```diff
@@ -40,8 +40,8 @@ object GoodsRecord {
       getCommodityCode(answers),
       answers.getPageValue(CountryOfOriginPage),
       getGoodsDescription(answers)
-    ).parMapN((eori, traderReference, commodity, countryOfOrigin, goodsDescription) =>
-      GoodsRecord(eori, traderReference, commodity.commodityCode, countryOfOrigin, goodsDescription)
+    ).parMapN((eori, traderReference, commodityCode, countryOfOrigin, goodsDescription) =>
```
besscerule: I think this can be simplified to 
`.parMapN(GoodsRecord.apply)`

besscerule (changes requested):

app/models/GoodsRecord.scala
vinoocs: Done! Thanks :)

besscerule (approved):

PR #102, opened 2024-05-30
TGP 1073 Get Category Assessments

Ewan-James-Donovan:
https://jira.tools.tax.service.gov.uk/browse/TGP-1073

Ticket says to get category assessments from OTT and save them. 
- Ticket says this should be done when categorise now is pressed, but that is just a page link.
- I make the call when going to categorisation start instead, same effect.
- Uses the commodity query in user answers to call and then set CategorisationQuery.
- Redirect to journey recovery if they end up there without a commodity to categorise
- otherwise, build Nick's categorisation info model from the response from OTT and save it to user answers

Also, I got to enjoy learning about higher order functions :)

### DIFF
--- a/app/connectors/OttConnector.scala
+++ b/app/connectors/OttConnector.scala
@@ -18,42 +18,54 @@ package connectors
 
 import config.Service
 import models.Commodity
+import models.ott.response.OttResponse
 import play.api.Configuration
 import play.api.http.Status.{NOT_FOUND, OK}
-import play.api.libs.json.JsResult
+import play.api.libs.json.{JsResult, Reads}
 import uk.gov.hmrc.http.client.HttpClientV2
 import uk.gov.hmrc.http.{Authorization, HeaderCarrier, HttpResponse, NotFoundException, StringContextOps, UpstreamErrorResponse}
 
+import java.net.URL
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
 class OttConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit ec: ExecutionContext) {
 
   private val baseUrl: Service              = config.get[Service]("microservice.services.online-trade-tariff-api")
-  private def ottUrl(commodityCode: String) =
+  private def ottCommoditiesUrl(commodityCode: String) =
     url"$baseUrl/ott/commodities/$commodityCode"
 
-  def getCommodityCode(commodityCode: String)(implicit hc: HeaderCarrier): Future[Commodity] = {
+  private def ottGreenLanesUrl(commodityCode: String) =
+    url"$baseUrl/ott/goods-nomenclatures/$commodityCode"
 
-    val newHeaderCarrier = hc.copy(authorization = Some(Authorization("bearerToken")))
+  private def getFromOtt[T](commodityCode: String, urlFunc: String => URL, authToken: String)(implicit hc: HeaderCarrier, reads: Reads[T]): Future[T] = {
+    val newHeaderCarrier = hc.copy(authorization = Some(Authorization(authToken)))
 
     httpClient
-      .get(ottUrl(commodityCode))(newHeaderCarrier)
+      .get(urlFunc(commodityCode))(newHeaderCarrier)
       .execute[HttpResponse]
       .flatMap { response =>
         response.status match {
           case OK =>
             response.json
-              .validate[Commodity]
+              .validate[T]
               .map(result => Future.successful(result))
               .recoverTotal(error => Future.failed(JsResult.Exception(error)))
-          case _  =>
+          case _ =>
             Future.failed(UpstreamErrorResponse(response.body, response.status))
         }
       }
-      .recoverWith { case e: NotFoundException =>
+      .recoverWith { case _: NotFoundException =>
         Future.failed(UpstreamErrorResponse("", NOT_FOUND))
       }
+  }
+
+  def getCommodityCode(commodityCode: String)(implicit hc: HeaderCarrier): Future[Commodity] = {
+    getFromOtt[Commodity](commodityCode, ottCommoditiesUrl, "bearerToken")
+  }
+
+  def getCategorisationInfo(commodityCode: String)(implicit hc: HeaderCarrier): Future[OttResponse] = {
+    getFromOtt[OttResponse](commodityCode, ottGreenLanesUrl, "bearerToken")
   }
 
 }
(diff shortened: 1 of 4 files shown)


app/connectors/OttConnector.scala
```diff
@@ -18,42 +18,55 @@ package connectors
...
-    val newHeaderCarrier = hc.copy(authorization = Some(Authorization("bearerToken")))
+  private def getFromOtt[T](commodityCode: String, urlFunc: String => URL, authToken: String)(implicit
+    hc: HeaderCarrier,
+    reads: Reads[T]
+  ): Future[T] = {
+    val newHeaderCarrier = hc.copy(authorization = Some(Authorization(authToken)))
 
     httpClient
-      .get(ottUrl(commodityCode))(newHeaderCarrier)
+      .get(urlFunc(commodityCode))(newHeaderCarrier)
```
ChrisCookOC, line 48: like the use of higher-order functions here 🙂

it/test/connectors/OttConnectorSpec.scala
```diff
@@ -80,4 +82,114 @@ class OttConnectorSpec
...
+
+      wireMockServer.stubFor(
+        get(urlEqualTo(s"/ott/goods-nomenclatures/123456"))
+          .willReturn(serverError())
+      )
+
+      connector.getCategorisationInfo("123456").failed.futureValue
+    }
+
+    "must return a not found future when the server returns a not found" in {
```
ChrisCookOC, line 197: this is just returning a failed future, not a "not found future"

test/controllers/CategoryGuidanceControllerSpec.scala
```diff
@@ -17,17 +17,92 @@
...
+          bind[OttConnector].toInstance(mockOttConnector)
+        )
+        .build()
+
+      running(application) {
+        val request = FakeRequest(GET, routes.CategoryGuidanceController.onPageLoad().url)
+        val result  = route(application, request).value
+
+        status(result) mustEqual SEE_OTHER
+        redirectLocation(result).value mustEqual routes.JourneyRecoveryController.onPageLoad().url
```
ChrisCookOC, line 105: Could check in here that we don't call the ottConnector if there is no commodity query? Might be unnecessary here, just popped into mind since we are binding the connector anyway

it/test/connectors/OttConnectorSpec.scala
```diff
@@ -26,6 +26,8 @@ import play.api.inject.guice.GuiceApplicationBuilder
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.http.test.WireMockSupport
 
+import scala.io.Source
```
besscerule: unused

app/connectors/OttConnector.scala
```diff
@@ -18,42 +18,55 @@ package connectors
...
       .flatMap { response =>
         response.status match {
           case OK =>
             response.json
-              .validate[Commodity]
+              .validate[T]
               .map(result => Future.successful(result))
               .recoverTotal(error => Future.failed(JsResult.Exception(error)))
           case _  =>
             Future.failed(UpstreamErrorResponse(response.body, response.status))
```
besscerule: can we write a test to check this

app/connectors/OttConnector.scala
```diff
@@ -18,42 +18,55 @@ package connectors
...
+      .get(urlFunc(commodityCode))(newHeaderCarrier)
       .execute[HttpResponse]
       .flatMap { response =>
         response.status match {
           case OK =>
             response.json
-              .validate[Commodity]
+              .validate[T]
               .map(result => Future.successful(result))
               .recoverTotal(error => Future.failed(JsResult.Exception(error)))
```
besscerule, line 56: can we write a test for this

besscerule (dismissed):

app/connectors/OttConnector.scala
Ewan-James-Donovan: That method is private but the public usages of this higher order function are tested: getCommodityCode, and getCategorisationInfo

app/connectors/OttConnector.scala
Ewan-James-Donovan: That method is private but the public usages of this higher order function are tested: getCommodityCode, and getCategorisationInfo

test/controllers/CategoryGuidanceControllerSpec.scala
Ewan-James-Donovan: Done :)

it/test/connectors/OttConnectorSpec.scala
Ewan-James-Donovan: Done :)

app/connectors/OttConnector.scala
Ewan-James-Donovan: If you're concerned about the upstream error, recall that there is default error handling and we shouldn't have to explicitly handle it (unless I'm mistaken)

app/connectors/OttConnector.scala
Ewan-James-Donovan: If you're concerned about the upstream error, recall that there is default error handling and we shouldn't have to explicitly handle it (unless I'm mistaken)

besscerule: it's just for code coverage, same for the one below, we don't have a test that hits that line

app/connectors/OttConnector.scala
besscerule: see above

app/connectors/OttConnector.scala
besscerule: i suppose you could argue if we don't have a test to hit it why is it there?

Ewan-James-Donovan: Good point, I'll have a think

Ewan-James-Donovan: Amended it, it was working wrong anyway. Added in a test, coverage should be fine now.

app/connectors/OttConnector.scala
```diff
@@ -18,42 +18,56 @@ package connectors
...
 import models.Commodity
+import models.ott.response.OttResponse
 import play.api.Configuration
-import play.api.http.Status.{NOT_FOUND, OK}
-import play.api.libs.json.JsResult
+import play.api.http.Status.{INTERNAL_SERVER_ERROR, NOT_FOUND, OK}
+import play.api.libs.json.{JsResult, Reads}
 import uk.gov.hmrc.http.client.HttpClientV2
-import uk.gov.hmrc.http.{Authorization, HeaderCarrier, HttpResponse, NotFoundException, StringContextOps, UpstreamErrorResponse}
+import uk.gov.hmrc.http.{Authorization, HeaderCarrier, HttpResponse, InternalServerException, NotFoundException, StringContextOps, Upstream5xxResponse, UpstreamErrorResponse}
```
ChrisCookOC: InternalServerException is unused

Ewan-James-Donovan: :)

ChrisCookOC (approved):

yohan-oc (approved):

vinoocs (approved):

PR #115, opened 2024-06-04
TGP-1139 create categorisation result page [Content only]

Ewan-James-Donovan:
This is the content only. There is a categorisation helper that will eventually decide with view to show. The view has some conditional logic.

### DIFF
--- a/app/models/helper/CategorisationHelper.scala
+++ b/app/models/helper/CategorisationHelper.scala
@@ -20,10 +20,9 @@ import models.UserAnswers
 
 object CategorisationHelper {
 
-  def getScenario(userAnswers: UserAnswers): Scenario = {
+  def getScenario(userAnswers: UserAnswers): Scenario =
     //TODO: Implement logic
     Category1NoExemptions
-  }
 
 }
 
--- a/app/models/helper/CategorisationHelper.scala
+++ b/app/models/helper/CategorisationHelper.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


app/models/helper/CategorisationHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package models.helper
+
+import models.UserAnswers
+
+object CategorisationHelper {
+
+  def getScenario(userAnswers: UserAnswers): Scenario =
+    //TODO: Implement logic
```
Ewan-James-Donovan, line 24: This is to be done when the ticket is picked back up.

app/views/CategorisationResultView.scala.html
```diff
@@ -14,14 +14,65 @@
...
+
+@panel = @{
+    val panelContent = scenario match {
+        case Category1NoExemptions => messages("categorisationResult.Category1NoExemptions.panelBody")
+        case Category3NoAssessments => messages("categorisationResult.Category3NoAssessments.panelBody")
+        case StandardNoSupplementaryUnits => messages("categorisationResult.StandardNoSupplementaryUnits.panelBody")
+        case Category1 => messages("categorisationResult.Category1.panelBody")
+        case Category2 => messages("categorisationResult.Category2.panelBody")
+    }
+    govukPanel(Panel(title = messages("categorisationResult.heading"), content = HtmlContent("<p class=\"govuk-body\" style=\"color:#FFFFFF\">" + panelContent + "</p>")))
```
Ewan-James-Donovan, line 35: The colour style needs to be here. govukPanel is made to the design system spec, and the prototype did not follow that exactly. Though the design system says nothing against what has been done. Anyway, this needs to be here otherwise other styling is used. Will not remove unless a better option is provided that works. I've already tried a few things

test/controllers/CategorisationResultControllerSpec.scala
```diff
@@ -37,7 +38,7 @@ class CategorisationResultControllerSpec extends SpecBase {
         val view = application.injector.instanceOf[CategorisationResultView]
 
         status(result) mustEqual OK
-        contentAsString(result) mustEqual view()(request, messages(application)).toString
+        contentAsString(result) mustEqual view(Category1NoExemptions)(request, messages(application)).toString
```
Ewan-James-Donovan, line 41: View now takes a scenario.

ChrisCookOC (approved):

mycp98 (approved):

PR #117, opened 2024-06-04
TGP-1016: post goods record

besscerule:


Ewan-James-Donovan (dismissed):

app/models/GoodsRecord.scala
```diff
@@ -26,7 +26,9 @@ final case class GoodsRecord(
   traderReference: String,
   commodityCode: String,
   countryOfOrigin: String,
-  goodsDescription: String
+  goodsDescription: String,
+  category: String,
```
ChrisCookOC: Does the category have to be there? When we do a Create Goods Record it will be before any categorisation has happened so we'll not have a category at that point. Maybe an optional?

app/models/GoodsRecord.scala
```diff
@@ -39,7 +41,10 @@ object GoodsRecord {
       answers.getPageValue(TraderReferencePage),
       getCommodityCode(answers),
       answers.getPageValue(CountryOfOriginPage),
-      getGoodsDescription(answers)
+      getGoodsDescription(answers),
+      // TODO remove hard coding
+      Right("1"),
+      Right("1970-01-01")
```
ChrisCookOC: This could be grabbed from the 

userAnswers.get(CommodityQuery) 

which has the effective date inside it

app/models/GoodsRecord.scala
besscerule: So this is a stopgap whilst we work out what we are doing with category - it might be case it. hardcoded as 0 on the router or something, or it's optional, for the moment it's here just to get things working

app/models/GoodsRecord.scala
besscerule: this took a LOT more work than I thought it would >_<

app/connectors/GoodsRecordConnector.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  private val tgpRouterBaseUrl: Service = config.get[Service]("microservice.services.trader-goods-profiles-router")
+  private val goodsRecordUrl            = url"$tgpRouterBaseUrl/trader-goods-profiles-router/records"
+
+  def submitGoodsRecordUrl(goodsRecord: CreateGoodsRecordRequest, eori: String)(implicit
+    hc: HeaderCarrier
+  ): Future[CreateGoodsRecordResponse] =
+    httpClient
+      .post(goodsRecordUrl)
+      .setHeader(header = ("X-Client-ID", "tss"))
```
besscerule: What goes in here? eori?

app/controllers/CyaCreateRecordController.scala
```diff
@@ -63,9 +68,31 @@ class CyaCreateRecordController @Inject() (
...
+              .map { _ =>
+                cleanseUserAnswers(request.userAnswers)
+                Redirect(routes.CreateRecordSuccessController.onPageLoad(goodsRecordResponse.recordId))
+              }
+          }
+      case Left(errors) => Future.successful(logErrorsAndContinue(errors))
+    }
+  }
+
+  def cleanseUserAnswers(userAnswers: UserAnswers): Done = {
```
besscerule: can't get this to work - helpful answers welcome

app/connectors/GoodsRecordConnector.scala
liandra-giacomine: This is added by the platform, not needed

app/controllers/CyaCreateRecordController.scala
```diff
@@ -63,9 +68,31 @@ class CyaCreateRecordController @Inject() (
   }
 
   def onSubmit(): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
-    auditService
-      .auditFinishCreateGoodsRecord(request.eori, request.affinityGroup, request.userAnswers)
-      .map(_ => Redirect(routes.CreateRecordSuccessController.onPageLoad().url))
+    CreateGoodsRecordRequest.build(request.userAnswers, request.eori) match {
+      case Right(model) =>
+        goodsRecordConnector
```
liandra-giacomine: is it possible to put this in a for comprehension? easier to process

app/controllers/CyaCreateRecordController.scala
```diff
@@ -63,9 +68,31 @@ class CyaCreateRecordController @Inject() (
...
+          .submitGoodsRecordUrl(model, request.eori)
+          .flatMap { goodsRecordResponse =>
+            auditService
+              .auditFinishCreateGoodsRecord(request.eori, request.affinityGroup, request.userAnswers)
+              .map { _ =>
+                cleanseUserAnswers(request.userAnswers)
```
liandra-giacomine: side effect (2 actions in map)

app/controllers/CyaCreateRecordController.scala
liandra-giacomine: Why are we returning Done?

app/models/CreateGoodsRecordRequest.scala
```diff
@@ -20,27 +20,52 @@ import cats.data.{EitherNec, NonEmptyChain}
...
       Right(eori),
       answers.getPageValue(TraderReferencePage),
-      getCommodityCode(answers),
       answers.getPageValue(CountryOfOriginPage),
-      getGoodsDescription(answers)
-    ).parMapN(GoodsRecord.apply)
+      getGoodsDescription(answers),
+      getCommodity(answers)
+    ).parMapN((eori, traderReference, countryOfOrigin, goodsDescription, commodity) =>
+      CreateGoodsRecordRequest(
```
liandra-giacomine: we want two separate objects, one for GoodsRecord and another for the request  - the request

app/connectors/GoodsRecordConnector.scala
Ewan-James-Donovan: I needed it a moment ago on some separate work, it doesn't seem to get auto populated for me at least. It might be in this case though.

besscerule: it's required for the router - they validate if the header is present

app/controllers/CyaCreateRecordController.scala
besscerule: because i've no idea what i am doing >_<

app/models/CreateGoodsRecordRequest.scala
besscerule: i was thinking about this - i guess the proper good record object would be created with the get goods record ticket - but i don't need it for mine so i didn't include it

app/connectors/GoodsRecordConnector.scala
Ewan-James-Donovan: roger that

app/models/CreateGoodsRecordRequest.scala
Ewan-James-Donovan: Exactly, I will pick up from here on the response and see if it fits in nicely - for now, great

app/controllers/CreateRecordSuccessController.scala
```diff
@@ -33,7 +33,8 @@ class CreateRecordSuccessController @Inject() (
 ) extends FrontendBaseController
     with I18nSupport {
 
-  def onPageLoad: Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
-    Ok(view())
+  def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request =>
+      Ok(view())
```
Ewan-James-Donovan: We can pass the recordId down to the view for link "Categorise now". This is good for now

conf/application.conf
```diff
@@ -47,7 +47,7 @@ microservice {
       trader-goods-profiles-router {
         protocol = http
         host     = localhost
-        port     = 10903
+        port     = 10904
```
Ewan-James-Donovan, line 50: I did this, too - we should probably all catch up about whether we point to router or stubs from here on out.

Ewan-James-Donovan (dismissed):
Will reapprove if needed, at the high level this is great.

Ewan-James-Donovan (dismissed):

it/test/connectors/GoodsRecordConnectorSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    "must submit a goods record" in {
+
+      val goodsRecordResponse = CreateGoodsRecordResponse("recordId")
+
+      wireMockServer.stubFor(
+        post(urlEqualTo(s"/trader-goods-profiles-router/records"))
+          .withRequestBody(equalTo(Json.toJson(goodsRecordRequest).toString))
+          .willReturn(ok().withBody(Json.toJson(goodsRecordResponse).toString))
```
ChrisCookOC, line 86: Since we now have code to add a header we could check that in the test

`.withHeader("X-Client-ID, "tss")`

or whatever into the test to ensure its being added

app/controllers/CyaCreateRecordController.scala
```diff
@@ -63,9 +65,14 @@ class CyaCreateRecordController @Inject() (
...
 
   def onSubmit(): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
-    auditService
-      .auditFinishCreateGoodsRecord(request.eori, request.affinityGroup, request.userAnswers)
-      .map(_ => Redirect(routes.CreateRecordSuccessController.onPageLoad().url))
+    CreateGoodsRecord.build(request.userAnswers, request.eori) match {
+      case Right(model) =>
+        for {
+          goodsRecordResponse <- goodsRecordConnector.submitGoodsRecord(model)
+          _                   <- auditService.auditFinishCreateGoodsRecord(request.eori, request.affinityGroup, request.userAnswers)
```
besscerule: this needs to be a <- because it is designed to fail if the audit doesn't work (see other controllers)

app/models/GoodsRecord.scala
besscerule: changed to be set in the request automatically

app/controllers/CyaCreateRecordController.scala
besscerule: removed and will implement in another ticket

app/controllers/CyaCreateRecordController.scala
besscerule: done but see comment

app/controllers/CyaCreateRecordController.scala
besscerule: removed

app/models/CreateGoodsRecordRequest.scala
besscerule: added in the router stuff, but still have my own separate create

app/controllers/CreateRecordSuccessController.scala
besscerule: done

conf/application.conf
besscerule: agreed

it/test/connectors/GoodsRecordConnectorSpec.scala
besscerule: will do in another ticket

app/controllers/CyaCreateRecordController.scala
ChrisCookOC: (To be extra pedantic, the audit itself can fail (e.g. the auditservice can give us back an error response), it's only if the Future fails - e.g. connection issues etc. we'd end up failing here)

app/connectors/GoodsRecordConnector.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class GoodsRecordConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit
+  ec: ExecutionContext
+) {
+
+  private val tgpRouterBaseUrl: Service = config.get[Service]("microservice.services.trader-goods-profiles-router")
+  private val goodsRecordUrl            = url"$tgpRouterBaseUrl/trader-goods-profiles-router/records"
+
+  private def mapToCreateRecordRequest(goodsRecord: CreateGoodsRecord): CreateRecordRequest =
```
liandra-giacomine: can CreateGoodsRecord just be called CreateGoodsRecord? We want the same domain for all operations related to the record

besscerule: ???? :P i think you've got a typo somewhere i'm not sure what you mean:)

app/connectors/GoodsRecordConnector.scala
liandra-giacomine: should be updated to tgp-frontend as discussed :)

app/connectors/GoodsRecordConnector.scala
liandra-giacomine: Can we extract this function to the CreateRecordRequest object? An apply function that takes in the goods record and returns an instance of CreateRecordRequest?

app/controllers/CategoryGuidanceController.scala
```diff
@@ -44,21 +44,22 @@ class CategoryGuidanceController @Inject() (
...
+          val ottResponseFuture = ottConnector.getCategorisationInfo(commodity.commodityCode)
 
-        for {
-          goodsNomenclature  <- ottResponseFuture
-          categorisationInfo <- Future.fromTry(Try(CategorisationInfo.build(goodsNomenclature).get))
-          updatedAnswers     <- Future.fromTry(request.userAnswers.set(CategorisationQuery, categorisationInfo))
-          _                  <- sessionRepository.set(updatedAnswers)
-        } yield Ok(view())
+          for {
+            goodsNomenclature  <- ottResponseFuture
```
liandra-giacomine: can you just put this in the for comprehension instead of saving it to a val since it's not re-used

app/controllers/CyaCreateRecordController.scala
liandra-giacomine: should be an = rather than <- for audit as Chris mentioned, the audit is fire and forget

app/models/CreateGoodsRecord.scala
```diff
@@ -49,14 +64,22 @@ object GoodsRecord {
...
     }
 
-  private def getCommodityCode(answers: UserAnswers): EitherNec[ValidationError, String] =
+  private def getCommodity(answers: UserAnswers): EitherNec[ValidationError, Commodity] =
     answers.getPageValue(CommodityCodePage) match {
-      case Right(data)  =>
+      case Right(code)  =>
         answers.getPageValue(HasCorrectGoodsPage) match {
-          case Right(true)  => Right(data)
+          case Right(true)  => getCommodityQuery(answers, code)
```
liandra-giacomine: Ah I thought the commodity page was returning a string rather than commodity

app/connectors/GoodsRecordConnector.scala
```diff
@@ -0,0 +1,58 @@
...
+import uk.gov.hmrc.http._
+import uk.gov.hmrc.http.client.HttpClientV2
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class GoodsRecordConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit
+  ec: ExecutionContext
+) {
+  private val xClientId: String         = config.get[String]("headers.xClientId")
```
liandra-giacomine: how come we put headers in the config? does it change per environment?

app/connectors/GoodsRecordConnector.scala
liandra-giacomine: can CreateGoodsRecord just be called GoodsRecord? :')

app/connectors/GoodsRecordConnector.scala
besscerule: it doesn't - seemed like a good spot will chnage

app/connectors/GoodsRecordConnector.scala
besscerule: done and done

app/models/CreateGoodsRecord.scala
besscerule: Have a read through how the commodity object is gotten - you will see that is uses the page and the query and then checks they both have the same code

it/test/connectors/GoodsRecordConnectorSpec.scala
besscerule: done in this ticket >_<

app/controllers/CategoryGuidanceController.scala
besscerule: done

app/controllers/CyaCreateRecordController.scala
besscerule: done

besscerule: just removed the test that was thinking it should fail

liandra-giacomine (approved):

PR #121, opened 2024-06-05
TGP-1115: updated content

besscerule:


app/views/EmailView.scala.html
```diff
@@ -29,6 +29,8 @@
 
 @layout(pageTitle = title(form, messages("email.title"))) {
 
+    <h1 class="govuk-heading-l">@messages("email.heading")</h1>
```
vinoocs, line 32: now we added the heading here, need to hide the other label below - otherwise we'll have two labels? 

label = Label(content = messages("email.heading"), classes = "govuk-visually-hidden")

yohan-oc: Also `@import viewmodels.LabelSize._` can be removed

vinoocs (approved):

PR #123, opened 2024-06-05
TGP-1250 - store "has correct goods" boolean

ChrisCookOC:


app/models/UserAnswers.scala
```diff
@@ -73,7 +73,7 @@ final case class UserAnswers(
 
     updatedData.flatMap { d =>
       val updatedAnswers = copy(data = d)
-      page.cleanup(Some(value), updatedAnswers)
+      page.cleanup(Some(value), updatedAnswers, this)
```
liandra-giacomine, line 76: this?

ChrisCookOC: Yeah I don't like the "this", but we are inside the "old user answers" instance here and that is what we want to pass into this cleanup routine so we can do comparison between updated and old.

liandra-giacomine: Ah right, can we extract the clean up outside of the set? seems like a side effect/hidden functionality

liandra-giacomine (approved):

vinoocs (approved):

PR #127, opened 2024-06-06
TGP 1236: Store and retrieve category assessments

Ewan-James-Donovan:
Here's the overview. Don't be alarmed by the description length, I just wanted to be clear. It's not perfect but it seems to serve purpose and fairly easy to use:

###  Changes:

`GoodsRecordConnector`: (the router connector)
- Now has a `getRecord(eori: String, recordId: String)` which returns `GetGoodsRecordResponse`, which has the minimal info needed on it. Later on we can make it more detailed, but for this purpose it's fine. When we get to submission, we should consider aligning exactly with the router's models. Thanks @ChrisCookOC for this chunk.

`GetGoodsRecordResponse`:
- Used by `GoodsRecordConnector` to hold the response from the router, contains the relevant info we need to get category info: recordId, comcode, country of origin. (Note that country of origin has not been added to the `OttConnector` yet) Thanks @ChrisCookOC for this chunk.

`RecordCategorisationsQuery` & `RecordCategorisations` model:
- Holds a `Map[recordId, CategorisationInfo]` in user answers for easy access without accidental overwrites.
- This way we can get the relevant category info depending on what record we're actually answering for.

`CategorisationService`:
- This is the fun bit. It's a service that you can use on assessment journey controllers like `AssessmentController` where categorisation info for the particular recordId is needed.
- You pass it the recordId which would come from the url for the controller. `.requireCategorisation(request, recordId)`
- If we don't have `CategorisationInfo` in `RecordCategorisations` for the recordId given, it calls the router via `GoodsRecordConnector` to get the comcode and country code, calls OTT via `OttConnector` to get the `CategorisationInfo` and then saves that to the `Map[recordId, CategorisationInfo]` (mentioned earlier) in user answers.
- If we have categorisations already in `RecordCategorisations` for the recordId, it won't make a call to get them again.
- Couldn't use an action because header carrier is not within implicit scope for actions as far as I'm aware, couldn't get it to be in scope either.

Some files have been moved to more appropriate packages.

#### Example usage of service and resulting populated user answers on `AssessmentController`:

```scala
  def onPageLoad(mode: Mode, recordId: String, assessmentId: String): Action[AnyContent] =
    (identify andThen getData andThen requireData).async { implicit request => {

        categorisationService.requireCategorisation(request, recordId).flatMap[Result] { _ =>
          val optionalResult = for {
            recordQuery <- request.userAnswers.get(RecordCategorisationsQuery)
            categorisationInfo <- recordQuery.records.get(recordId)
            assessment <- categorisationInfo.categoryAssessments.find(_.id == assessmentId)
            assessmentIndex = categorisationInfo.categoryAssessments.indexOf(assessment)
          } yield {

... etc
```


app/models/router/responses/GetGoodsRecordResponse.scala
```diff
@@ -0,0 +1,50 @@
...
+object GetGoodsRecordResponse {
+
+  implicit val reads: Reads[GetGoodsRecordResponse] = {
+
+    import play.api.libs.functional.syntax._
+
+    (
+      (__ \ "recordId").read[String] and
+        (__ \ "comcode").read[String] and
+        (__ \ "countryOfOrigin").read[String]
```
Ewan-James-Donovan, line 36: Refer to PR description for why this is sparse.

app/services/CategorisationService.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
+
+class CategorisationService @Inject() (
+  sessionRepository: SessionRepository,
+  ottConnector: OttConnector,
+  goodsRecordsConnector: GoodsRecordConnector
+)(implicit ec: ExecutionContext) {
+
+  def requireCategorisation(request: DataRequest[_], recordId: String)(implicit hc: HeaderCarrier): Future[Done] = {
```
Ewan-James-Donovan, line 38: Refer to PR description for high level description of this service

app/models/RecordCategorisations.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import models.ott.CategorisationInfo
+import play.api.libs.json.{Json, OFormat}
+
+case class RecordCategorisations(records: Map[String, CategorisationInfo])
```
Ewan-James-Donovan, line 22: Refer to PR description how this is used

app/connectors/GoodsRecordConnector.scala
```diff
@@ -32,14 +34,35 @@ class GoodsRecordConnector @Inject() (config: Configuration, httpClient: HttpCli
...
+  def getRecord(eori: String, recordId: String)(implicit
+    hc: HeaderCarrier
+  ): Future[GetGoodsRecordResponse] =
+    httpClient
+      .get(getRecordUrl(eori, recordId))
+      .addHeaders(clientIdHeader)
+      .execute[HttpResponse]
+      .flatMap { response =>
+        response.status match {
+          case OK =>
```
liandra-giacomine: what happens when we don't get an OK? Will it be throwing an exception as it won't be a successful response? If so we can just parse the json in the flatmate and not worry about a match

you get the code below by default if you use response.json.as[GetGoodsRecordResponse]

.map(result => Future.successful(result))
               .recoverTotal(error => Future.failed(JsResult.Exception(error)))

Ewan-James-Donovan: Good point, I think that would work, it should fail the future with an exception in cases where things break anyway :)
I'll try it out

app/services/CategorisationService.scala
```diff
@@ -0,0 +1,62 @@
...
+      request.userAnswers.get(RecordCategorisationsQuery).getOrElse(RecordCategorisations(Map.empty))
+
+    recordCategorisations.records.get(recordId) match {
+      case Some(categorisationInfo: CategorisationInfo) =>
+        Future.successful(Done)
+      case None                                         =>
+        for {
+          goodsRecord        <- goodsRecordsConnector.getRecord(eori = request.eori, recordId = recordId)
+          goodsNomenclature  <- ottConnector.getCategorisationInfo(goodsRecord.commodityCode)
+          categorisationInfo <- Future.fromTry(Try(CategorisationInfo.build(goodsNomenclature).get))
```
liandra-giacomine, line 50: CategorisationInfo.build(goodsNomenclature).get

Are we catching an exception here if the get for an option fails?

app/services/CategorisationService.scala
```diff
@@ -0,0 +1,62 @@
...
+class CategorisationService @Inject() (
+  sessionRepository: SessionRepository,
+  ottConnector: OttConnector,
+  goodsRecordsConnector: GoodsRecordConnector
+)(implicit ec: ExecutionContext) {
+
+  def requireCategorisation(request: DataRequest[_], recordId: String)(implicit hc: HeaderCarrier): Future[Done] = {
+
+    val recordCategorisations =
+      request.userAnswers.get(RecordCategorisationsQuery).getOrElse(RecordCategorisations(Map.empty))
```
liandra-giacomine, line 41: So this does a getOrCreate type of thing?

liandra-giacomine (commented):
Looks really good, a couple of qs but functionality looks great

app/services/CategorisationService.scala
Ewan-James-Donovan: It's a future from try: Future.fromTry(Try(CategorisationInfo.build(goodsNomenclature).get)) so it'll just fail the future as a whole

app/services/CategorisationService.scala
Ewan-James-Donovan: Pretty much. It's a convenience thing. request.userAnswers.get(RecordCategorisationsQuery) would return Option[RecordCategorisations].  If it returned None, that would be logically equivalent to an empty map anyway. Means I can look in the map for a recordId, even when there isn't a map to look in yet.

app/connectors/GoodsRecordConnector.scala
Ewan-James-Donovan: Did something similarly simple now :)

app/services/CategorisationService.scala
liandra-giacomine: I'd say functionally, if it's a None then there's no need to look in the map?

liandra-giacomine (approved):

ChrisCookOC (approved):

PR #129, opened 2024-06-06
TGP-1149: update goods record partially finished

besscerule:


app/models/UserAnswers.scala
```diff
@@ -37,7 +37,7 @@ final case class UserAnswers(
   def getPageValue[A](page: Gettable[A])(implicit rds: Reads[A]): EitherNec[ValidationError, A] =
     get(page).map(Right(_)).getOrElse(Left(NonEmptyChain.one(PageMissing(page))))
 
-  def getOptionalPageValue(
+  def getOptionalPageStringValue(
```
ChrisCookOC: Can do this with typing and so not need to define the Int one e.g.

`    
  def getOptionalPageValue[T](answers: UserAnswers,
    questionPage: QuestionPage[Boolean],
    optionalPage: QuestionPage[T]
  ): EitherNec[ValidationError, Option[T]] =
    getPageValue(questionPage) match {
      case Right(true)  => getPageValue(optionalPage).map(Some(_))
      case Right(false) => unexpectedValueDefined(answers, optionalPage)
    }`

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,57 @@
...
+)
+
+object CategoryRecord {
+
+  implicit lazy val format: OFormat[CategoryRecord] = Json.format
+
+  def build(answers: UserAnswers, eori: String, recordId: String): EitherNec[ValidationError, CategoryRecord] =
+    (
+      getCategory(answers),
+      answers.getOptionalPageIntValue(answers, HasSupplementaryUnitPage, SupplementaryUnitPage),
```
ChrisCookOC: Will this fail to build if they haven't answered the HasSupplementaryUnitPage? I am/was under the impression that we might not always show them that page

ChrisCookOC (commented):
Not fully reviewed yet, just a couple of things that caught my eye based on what I'm looking at currently :)

app/models/UserAnswers.scala
besscerule: agreed

app/models/CategoryRecord.scala
besscerule: at the moment it is a required page - but given the cat stuff is a tad up in the air this is what it is for the moment, but agreed it needs to be sorted

Ewan-James-Donovan (dismissed):
LGTM, will reapprove on any updates if they're safe

ChrisCookOC (dismissed):

Ewan-James-Donovan (commented):
LGTM, Discussed via slack about no assessments from OTT scenario. Well done :)) (This is an approval in theory)

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,101 @@
...
+        eori = eori,
+        recordId = recordId,
+        category = category,
+        supplementaryUnit = supplementaryUnit,
+        measurementUnit = measurementUnit
+      )
+    )
+
+  def getCategory(answers: UserAnswers, recordId: String): EitherNec[ValidationError, Int] =
+    answers.get(RecordCategorisationsQuery) match {
```
liandra-giacomine: you can use a map here instead of matching on a some

liandra-giacomine: answers.get(RecordCategorisationsQuery).map(recordCategorisations => recordCategorisations.records.get(recordId)).map(categorisationInfo => chooseCategory(recordId, answers, categorisationInfo))

or better, put it in a for comprehension

```
for {
recordCategorisations  <- answers.get(RecordCategorisationsQuery)
...
categorisationInfo <- chooseCategory(recordId, answers, categorisationInfo)
yield result 
```

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,101 @@
...
+    categorisationInfo: CategorisationInfo
+  ): EitherNec[ValidationError, Int] = {
+    val category1AssessmentsCount = categorisationInfo.categoryAssessments.count(_.category == 1)
+    if (category1AssessmentsCount > 0) {
+      answers.getPageValue(
+        AssessmentPage(recordId, category1AssessmentsCount - 1)
+      ) match {
+        case Right(AssessmentAnswer.NoExemption) => Right(1)
+        case Right(_)                            => chooseCategory2Or3(recordId, answers, categorisationInfo)
+        case _                                   => Right(1)
```
liandra-giacomine: This case would be a Left...seems confusing bc this would be a validation error but then we return a right

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,101 @@
...
+      }
+    } else {
+      chooseCategory2Or3(recordId, answers, categorisationInfo)
```
liandra-giacomine, line 106: can't have category 2 assessments without cat 1 I think? would this not then be a category 3?

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+  def chooseCategory2Or3(
+    recordId: String,
+    answers: UserAnswers,
+    categorisationInfo: CategorisationInfo
+  ): EitherNec[ValidationError, Int] = {
+    val category2AssessmentsCount = categorisationInfo.categoryAssessments.count(_.category == 2)
+    if (category2AssessmentsCount > 0) {
+      answers.getPageValue(AssessmentPage(recordId, categorisationInfo.categoryAssessments.length - 1)) match {
```
liandra-giacomine, line 81: Instead of using magic numbers, can we create enums for these? Associate these numbers to values such as CategoryOne, CategoryTwo and StandardGoods for cat 3.

app/models/CategoryRecord.scala
```diff
@@ -0,0 +1,112 @@
...
+        case Right(_)                            => chooseCategory2Or3(recordId, answers, categorisationInfo)
+        case _                                   => CATEGORY_1
+      }
+    } else {
+      chooseCategory2Or3(recordId, answers, categorisationInfo)
+    }
+  }
+
+  //TODO: get measurementUnit from answers
+  private def getMeasurementUnit(answers: UserAnswers): EitherNec[ValidationError, Option[String]] = Right(Some("1"))
```
liandra-giacomine, line 111: Will be updated once @yohan-oc  gets this bit done in his ticket to get this data from OTT, it will be more something like kilograms

liandra-giacomine (dismissed):

liandra-giacomine (approved):

PR #134, opened 2024-06-12
Tgp 372/implement dropdown of countries

besscerule:


mycp98 (approved):

app/controllers/CyaCreateRecordController.scala
```diff
@@ -41,29 +43,43 @@ class CyaCreateRecordController @Inject() (
...
-            UseTraderReferenceSummary.row(request.userAnswers),
-            GoodsDescriptionSummary.row(request.userAnswers),
-            CountryOfOriginSummary.row(request.userAnswers),
-            CommodityCodeSummary.row(request.userAnswers)
-          ).flatten
-        )
-        Ok(view(list))
-      case Left(errors) => logErrorsAndContinue(errors)
+        request.userAnswers.get(CountriesQuery) match {
+          case Some(countries) => Future.successful(displayView(request.userAnswers, countries))
```
liandra-giacomine, line 58: how come we need to retrieve the country list for CYA? Can't we just retrieve the CountryOfOriginPage value or am I oversimplifying it?

besscerule: No because we need to check that it is from the required list of countries for error checking

liandra-giacomine (approved):

app/models/Country.scala
```diff
@@ -0,0 +1,61 @@
...
+import viewmodels.govuk.select._
+
+final case class Country(
+  id: String,
+  description: String
+)
+
+object Country {
+
+  def selectItems(countries: Seq[Country])(implicit messages: Messages): Seq[SelectItem] =
```
ChrisCookOC, line 31: Need them to display sorted by name rather than Country Code

PR #137, opened 2024-06-13
TGP-1139: link up cat result page

besscerule:


vinoocs:
While testing in local, found below -not sure if this is expected though. 

Scenario 5
**Expected:**
And **I have provided document codes for category 1 assessments but none for category 2 assessments**
**And my record has been categorised as category 2**
Then I should be navigated to [category confirmation page - result category 2](https://trader-goods-profile-prototype-18527416d281.herokuapp.com/version-4/categorisation-supp-units/categorisation-result-supp-units)

But when I provide document codes for cat1 & none for cat 2 it still shows it as **Category 1**
image

image

Also, 
Scenario 6

When I select Go to homepage
Then I should be navigated to /homepage
But I am not navigated to homepage

app/models/Scenario.scala
```diff
@@ -14,22 +14,32 @@
...
+case object StandardNoAssessments extends Scenario
+case object Standard extends Scenario
 case object Category1 extends Scenario
 case object Category2 extends Scenario
+
+object Scenario {
+
+  def getScenario(goodsRecord: CategoryRecord): Scenario =
+    goodsRecord.category match {
+      case 1 => Category1
```
vinoocs, line 33: While testing in local, found below, Please correct me if this is expected 

Scenario 5
**Expected:**
And **I have provided document codes for category 1 assessments but none for category 2 assessments**
**And my record has been categorised as category 2**
Then I should be navigated to [category confirmation page - result category 2](https://trader-goods-profile-prototype-18527416d281.herokuapp.com/version-4/categorisation-supp-units/categorisation-result-supp-units)

But when I provide document codes for cat1 & none for cat 2 it still shows it as **Category 1** instead of **Category 2**
image

image

Also, 
Scenario 6

When I select Go to homepage
Then I should be navigated to /homepage
**But I am not navigated to homepage**

besscerule: will check this out - thanks for the detailed comment

liandra-giacomine: it depends on whether assessment 2 is for cat 1 or cat 2

liandra-giacomine: @vinoocs what commodity code did you use? :)

besscerule: From having looked at some examples this is expected behaviour. The cat 2 of 6 doesn't mean cat 2 - that could be a cat 1 good question in this case - so it should be cat 1 goods.

vinoocs: @liandra-giacomine I used the commodity code - 1601001011

ChrisCookOC (dismissed):

ChrisCookOC (approved):

vinoocs (approved):

PR #140, opened 2024-06-14
TGP-189 - Create Supplementary Unit Page

yohan-oc:


app/controllers/SupplementaryUnitController.scala
```diff
@@ -53,15 +54,31 @@ class SupplementaryUnitController @Inject() (
         case Some(value) => form.fill(value)
       }
 
-      Ok(view(preparedForm, mode, recordId))
+      request.userAnswers.get(RecordCategorisationsQuery) match {
+        case None        => Redirect(routes.JourneyRecoveryController.onPageLoad().url)
+        case Some(query) =>
+          query.records(recordId).measurementUnit match {
```
ChrisCookOC: not sure we can/should be assuming
query.records(recordId)
is defined?

Could be worth switching to a for/yield like in the other places we get the details out of the recordCategorisationsQuery

app/controllers/SupplementaryUnitController.scala
```diff
@@ -53,15 +54,31 @@ class SupplementaryUnitController @Inject() (
...
+            case None                  => Ok(view(preparedForm, mode, recordId, ""))
```
ChrisCookOC: Do we want to load the page if there is no measurementUnit?

test-utils/generators/SupplementaryUnitAmountGenerator.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def formatDoubleWithCommas(value: Double): String = {
+    val formatted   = f"$value%.6f" // Format with 6 decimal places
+    val parts       = formatted.split("\\.")
+    val integerPart = parts(0).reverse.grouped(3).mkString(",").reverse // Add commas every three digits
+    val decimalPart = if (parts.length > 1) "." + parts(1) else "" // Append decimal part if present
+    integerPart + decimalPart
+  }
+
+  def genDoublesperseString(gen: Gen[String], value: String, frequencyV: Int = 1, frequencyN: Int = 10): Gen[String] = {
```
ChrisCookOC: what does Doublesperse mean?

app/views/SupplementaryUnitView.scala.html
```diff
@@ -37,10 +39,11 @@
         @govukInput(
             InputViewModel(
                 field = form("value"),
-                label = LabelViewModel(messages("supplementaryUnit.heading")).asPageHeading()
+                label = Label(content = "supplementaryUnit.heading", classes = "govuk-visually-hidden")
             )
             .asNumeric()
-            .withWidth(Fixed10)
+            .withWidth(Fixed5)
```
ChrisCookOC: This isn't big enough to support the -9999999999.999999 we support?

app/controllers/SupplementaryUnitController.scala
```diff
@@ -53,15 +54,31 @@ class SupplementaryUnitController @Inject() (
...
+      }
   }
 
   def onSubmit(mode: Mode, recordId: String): Action[AnyContent] =
     (identify andThen getData andThen requireData).async { implicit request =>
       form
         .bindFromRequest()
         .fold(
-          formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, recordId))),
+          formWithErrors =>
```
ChrisCookOC: Some formatting stuff happens with big numbers

Enter 1234567890.123456
image

It turns it into scientific notation
image

Click Change

it is still in scientific notation
image

Without changing the value, if I click continue its unhappy with the format
image

Should probably ensure its displayed in full form everytime?

test-utils/generators/SupplementaryUnitAmountGenerator.scala
yohan-oc: Name supposed to be gen**Intersperse**String. This is now unused method. Removing this :)

ChrisCookOC: Oh haha. Find and Replace strikes again 😆

app/views/SupplementaryUnitView.scala.html
yohan-oc: This is according to Prototype. https://trader-goods-profile-prototype-18527416d281.herokuapp.com/version-4/categorisation-longer-code-supp-units/supplementary-unit-longer-supp
 
 I will raise this to design team

test-utils/generators/SupplementaryUnitAmountGenerator.scala
yohan-oc: facepalm!

app/views/SupplementaryUnitView.scala.html
yohan-oc: Design team agreed to increase the textbox length. Changed back to `.withWidth(Fixed10)`

app/controllers/SupplementaryUnitController.scala
yohan-oc: This is now fixed. I have changed formProvider from double to String

app/controllers/SupplementaryUnitController.scala
yohan-oc: Updated to for/yield

app/controllers/SupplementaryUnitController.scala
yohan-oc: Yes. measurementUnit can be null https://api.trade-tariff.service.gov.uk/green-lanes.html#get-green-lanes-goods-nomenclatures-id

ChrisCookOC (dismissed):

ChrisCookOC (approved):

vinoocs (approved):

PR #141, opened 2024-06-17
TGP-1183: Create Longer commodity code input page

mycp98:


app/controllers/LongerCommodityCodeController.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def onPageLoad(mode: Mode, recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request =>
+      val preparedForm = request.userAnswers.get(LongerCommodityCodePage) match {
+        case None        => form
+        case Some(value) => form.fill(value)
+      }
+
+      request.userAnswers.get(CommodityCodePage) match {
+        case Some(commodity) => Ok(view(preparedForm, mode, commodity, recordId))
```
ChrisCookOC: Do we still want to load this page/view if the commodity code is already 10 digits? Guess we would in practice handle this before we get here (e.g. never loading this page if 10 digit commodity code)

mycp98: i've updated this now, thanks

ChrisCookOC (approved):

vinoocs (approved):

PR #142, opened 2024-06-17
TGP-398: Create previous movement records page

vinoocs:
- Used HTML tags for loadspinner as we do not have gov component at the moment
- JS & CSS added for event & styling on load spinner
- Currently, we fetch the records from B&T through router and store them in data-store and also store in frontend Session(userAnswers)
- Storing the flag in session to indicate the user requested loading of records
- The records have been stored in the GetGoodsRecordsQuery


app/connectors/GetGoodsRecordsConnector.scala
```diff
@@ -0,0 +1,53 @@
...
+
+class GetGoodsRecordsConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit
+  ec: ExecutionContext
+) {
+
+  private val dataStoreBaseUrl: Service = config.get[Service]("microservice.services.trader-goods-profiles-data-store")
+  private val clientIdHeader            = ("X-Client-ID", "tgp-frontend")
+  private def getGoodsRecordsUrl(
+    eori: String,
+    lastUpdatedDate: Option[String] = None,
```
ChrisCookOC: Probably no need to default these in, they are all in scope in the call below

app/models/router/responses/GoodsItemRecords.scala
```diff
@@ -0,0 +1,118 @@
...
+ * limitations under the License.
+ */
+
+package models.router.responses
+
+import play.api.libs.json._
+import utils.ResponseModelSupport.removeNulls
+
+import java.time.Instant
+case class GoodsItemRecords(
```
ChrisCookOC, line 23: Do we need all these fields for our purposes or just a subset?

app/connectors/GetGoodsRecordsConnector.scala
```diff
@@ -0,0 +1,53 @@
...
+import config.Service
+import models.router.responses.GetRecordsResponse
+import play.api.Configuration
+import uk.gov.hmrc.http._
+import uk.gov.hmrc.http.client.HttpClientV2
+
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class GetGoodsRecordsConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit
```
ChrisCookOC: scope-wise, this could probably slot into the existing GoodsRecordsConnector nicely

app/connectors/GetGoodsRecordsConnector.scala
```diff
@@ -0,0 +1,53 @@
...
+  private val clientIdHeader            = ("X-Client-ID", "tgp-frontend")
+  private def getGoodsRecordsUrl(
+    eori: String,
+    lastUpdatedDate: Option[String] = None,
+    page: Option[Int] = None,
+    size: Option[Int] = None
+  )                                     =
+    url"$dataStoreBaseUrl/trader-goods-profiles-data-store/traders/$eori/records?lastUpdatedDate=$lastUpdatedDate&page=$page&size=$size"
+
+  def getRecords(
```
ChrisCookOC: Do we ever call this with the optional params?

app/models/router/responses/GoodsItemRecords.scala
vinoocs: Probably be, @liandra-giacomine - please can you confirm.

app/connectors/GetGoodsRecordsConnector.scala
vinoocs: Yes, makes sense. Moved it to GoodsRecordsConnector

app/connectors/GetGoodsRecordsConnector.scala
vinoocs: Good catch, @liandra-giacomine - what are the default values we can set for lastUpdatedDate, page and size here.

app/connectors/GetGoodsRecordsConnector.scala
vinoocs: Done it!

app/connectors/GetGoodsRecordsConnector.scala
vinoocs: Realised we do not need these optional params here so removed them. The reason I had was, the getRecords was always returning only one record if we don't provide optional params. The bug has been raised by @bethan.dobb to the API team so hopefully it gets fixed.

liandra-giacomine (approved):

app/models/router/responses/GoodsItemRecords.scala
vinoocs: HAd a chat with @liandra-giacomine - we need to refactor a few things as part of TGP-1061 and TGP-1221 so will take care of the changes in that ticket.

ChrisCookOC (approved):

PR #144, opened 2024-06-17
TGP-1434: Updated OTT url config

liandra-giacomine:
Updated the config for OTT from a Service (host, port) to a url so that we can easily switch from the stub to the actual OTT endpoints in QA and PROD


yohan-oc (commented):
There are some merge confilicts

ChrisCookOC (dismissed):

mycp98 (approved):

PR #147, opened 2024-06-18
TGP-1322 TGP-1323 audit OTT calls

ChrisCookOC:
**Overarching thoughts**
Tried to keep it using the same flow as before where there is just one bit of code doing all the call to OTT and using types and a new enum ~~first-order functions~~  to keep it that way. Hopefully it is clear what is happening.

**OttAuditDetails**
This class is to collect together things, partially to make it easier to pass them around, partially to make the ~~first-order~~ "which event is being audited" function definition easier, and also partially to stop intellij banging on about passing too many parameters into functions.

**Recover with**
Part of the scope of these audits are to report errors as well. So needed to catch the errors, audit them and then rethrow them. Not sure its ideal but its there. Split into http-y ones from the httpclient that we can get a nice status from, and then just general ones from e.g. json-y nonsense

**Extended Data Event vs Data Event**
In the past audit events we've used a DataEvent, which takes a map of string keys and string values. This is fine for generating a nice flat JSON
However, for these events we want a nested structure. This requires using a Json object and an ExtendedAuditEvent.

**null**
On some of the fields in the tickets and in other audit events its been proposed we pass "null" into fields when empty. For consistency I've continued that on all the optional fields. But maybe it will change when Kara reviews. Since we have to have a json object (see above), it would be easy and nice to just use options and not care when they are empty. Will see what Kara thinks when she is back.

**Http status code descriptions** 
Couldn't find a library function or whatever to do this so I stole a map from another HMRC project.

**Date of Trade** 
Was asked to audit this, no clue where it comes from. Hardcoded to today for now.


liandra-giacomine (changes requested):
I don't understand why we would pass the audit function as an input parameter instead of calling it directly in the connector? It seems to add unnecessary complexity?

app/connectors/OttConnector.scala
```diff
@@ -42,38 +47,113 @@ class OttConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(i
...
+      None,
+      None,
+      Some(journey)
+    )
 
-  def getCommodityCode(commodityCode: String)(implicit hc: HeaderCarrier): Future[Commodity] =
-    getFromOttWithCommodityCode[Commodity](commodityCode, ottCommoditiesUrl, "bearerToken")
+    getFromOtt[Commodity](
+      ottCommoditiesUrl(commodityCode),
+      "bearerToken",
```
liandra-giacomine: can we make the auth token an environment variable? I need to set OTT's staging token for it to work in QA so can't be a hard-coded string we pass in

app/connectors/OttConnector.scala
```diff
@@ -42,38 +47,113 @@ class OttConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(i
...
+
+          Future.failed(UpstreamErrorResponse(e.message, e.responseCode))
+
+        case e: UpstreamErrorResponse =>
+          auditService.auditOttCall(auditDetails, requestStartTime, Instant.now, e.statusCode, Some(e.message), None)
+          Future.failed(e)
+
+        case e: Exception =>
+          //E.g. Any error not directly related to the http call, e.g. Json parsing
+          auditService.auditOttCall(auditDetails, requestStartTime, Instant.now, OK, Some(e.getMessage), None)
```
liandra-giacomine: I'm not sure about auditing in this case, the status OK doesn't seem to make sense here

liandra-giacomine (commented):
A cheeky request and a question but it's looking awesome

app/connectors/OttConnector.scala
ChrisCookOC: Done

app/connectors/OttConnector.scala
ChrisCookOC: I realised this case was only really intended to catch that error we throw if we got back a successful OK from the server but couldn't parse the json. So I've just moved the audit to there and ditched this case.

liandra-giacomine: Awesome :)

liandra-giacomine (dismissed):

liandra-giacomine (approved):

yohan-oc (approved):

PR #161, opened 2024-06-26
TGP-1518: updated content

vinoocs:
- The content has been updated
- The regex & char limit logic already exists and is appropriate so no changes to that. 
- Journey test has been run to make sure it doesn't fail just in case


yohan-oc (approved):
LGTM

mycp98 (approved):

PR #166, opened 2024-06-28
TGP-1140: Implement navigation based on assessments/exemptions

Ewan-James-Donovan:
Other pr: https://github.com/hmrc/trader-goods-profiles-stubs/pull/66

https://jira.tools.tax.service.gov.uk/browse/TGP-1140

This ticket basically is to cover these cases after clicking "Categorise Goods Now" or otherwise starting the categorisation journey:

1) There are category 1 assessments that have no exemptions -> `Category1NoExemptions` -> Redirect to `CreateRecordSuccess` and update goodsRecord with cat 1.
2) There are no category assessments whatsoever -> `StandardNoAssessments` -> Redirect to `CreateRecordSuccess` and update goodsRecord with cat 3.
3) Otherwise -> `NoRedirectScenario` -> stay on categorisation start page and continue as normal

In lieu of creating a controller solely for redirecting, I have refactored `CategorisationStartController` to decide whether to redirect. I have also removed code from there and a test which predated our `recordId` way of updating.

You'll notice I've added a scenario called `NoRedirectScenario`, and a new method on `Scenario` called `getRedirectScenarios`. The method will return `NoRedirectScenario` for case (3), otherwise, one of the other two will be returned. `NoRedirectScenario` simply symbolises that we're going to continue as normal and nothing more.

### DIFF
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -54,7 +54,7 @@ class CategoryGuidanceController @Inject() (
           val scenario = Scenario.getScenario(categoryRecord)
           val isRedirectScenario = (scenario == StandardNoAssessments || scenario == Category1NoExemptions)
           if (isRedirectScenario) {
-            Future.successful(Redirect(routes.IndexController.onPageLoad.url))
+            Future.successful(Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario).url))
           } else {
             Future.successful(Ok(view(recordId)))
           }
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -52,8 +52,7 @@ class CategoryGuidanceController @Inject() (
         categorisationInfo <- Future.fromTry(Try(recordCategorisations.records.get(recordId).get))
         scenario = Scenario.getRedirectScenarios(recordCategorisations, categorisationInfo)
       } yield scenario match {
-        case Category1NoExemptions => Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario).url)
-        case StandardNoAssessments => Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario).url)
+        case Category1NoExemptions | StandardNoAssessments => Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario).url)
         case NoRedirectScenario => Ok(view(recordId))
       }).recover {
         case _ => Redirect(routes.JourneyRecoveryController.onPageLoad().url)
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -46,21 +46,21 @@ class CategoryGuidanceController @Inject() (
 
   def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData).async {
     implicit request =>
-          (for {
-            ua <- categorisationService.requireCategorisation(request, recordId)
-            categoryRecordEither = CategoryRecord.build(ua, request.eori, recordId)
-          } yield categoryRecordEither match {
-            case Right(categoryRecord) =>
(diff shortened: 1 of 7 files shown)


vinoocs (dismissed):
LGTM

app/controllers/CategoryGuidanceController.scala
```diff
@@ -38,21 +40,35 @@ class CategoryGuidanceController @Inject() (
...
+        scenario               = Scenario.getRedirectScenarios(categorisationInfo)
+        categoryRecord        <- Future.fromTry(Try(CategoryRecord.build(ua, request.eori, recordId).right.get))
+      } yield scenario match {
+        case Category1NoExemptions | StandardNoAssessments =>
+          goodsRecordConnector
+            .updateGoodsRecord(request.eori, recordId, categoryRecord)
+            .map { _ =>
+              Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario).url)
+            }
+            .recover { case _ =>
```
liandra-giacomine: you can put one recover block which encapsulates the whole for comprehension instead of two inside it 

Looks like we only need categoryRecord when it's Category1NoExemptions | StandardNoAssessments so we can do that in there, maybe separate the handling of different cases into a separate function for readability

app/models/Scenario.scala
```diff
@@ -28,6 +30,28 @@ case object Category2 extends Scenario
 
 object Scenario {
 
+  def getRedirectScenarios(categorisationInfo: CategorisationInfo): Scenario = {
+    val hasCategoryAssessments: Boolean =
+      categorisationInfo.categoryAssessments.length >= 1
```
liandra-giacomine: Is categoryAssessments a List? There's probably a scala native method for that, .nonEmpty or something?

app/controllers/CategoryGuidanceController.scala
```diff
@@ -38,21 +40,35 @@ class CategoryGuidanceController @Inject() (
...
-        case Some(_) =>
-          for {
-            _ <- categorisationService.requireCategorisation(request, recordId)
-          } yield Ok(view(recordId))
-        case None    =>
-          Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url))
-      }
+      (for {
+        ua                    <- categorisationService.requireCategorisation(request, recordId)
+        recordCategorisations <- Future.fromTry(Try(ua.get(RecordCategorisationsQuery).get))
```
liandra-giacomine: Future.fromTry(Try(ua.get(RecordCategorisationsQuery).get))

Let's catch up on how to handle this functionally instead of throwing an exception and catching it in a try :)

liandra-giacomine (commented):
Looking awesome, keen to pair a bit to refactor a couple of things

liandra-giacomine (changes requested):
Comments above :D

app/models/Scenario.scala
Ewan-James-Donovan: yep

app/controllers/CategoryGuidanceController.scala
```diff
@@ -38,21 +41,41 @@ class CategoryGuidanceController @Inject() (
...
+          val scenario              = categorisationInfo.map(Scenario.getRedirectScenarios)
+          scenario match {
+            case Some(Category1NoExemptions | StandardNoAssessments) =>
+              CategoryRecord
+                .build(userAnswers, request.eori, recordId)
+                .map { categoryRecord =>
+                  goodsRecordConnector
+                    .updateGoodsRecord(request.eori, recordId, categoryRecord)
+                    .map { _ =>
+                      Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario.get).url)
```
ChrisCookOC, line 67: Ideally should go in the navigator but will get swept up in a tech debt ticket anyway so 🤷

Ewan-James-Donovan: Thanks for letting me get away with it here, I'm onto something extra now anyway! haha

liandra-giacomine (approved):

PR #167, opened 2024-06-28
TGP-1061: new url for no records

vinoocs:
- A new url has been proposed by design for traders without records.
- Tested in local and works fine


Ewan-James-Donovan:
Nice and simple, LGTM

Ewan-James-Donovan (approved):

PR #168, opened 2024-06-28
TGP-1515 updated url's

AmarkumarJagtap:


vinoocs:
Looks good, @Amar - the journey test needs to be updated here https://github.com/hmrc/trader-goods-profiles-ui-journey-tests/blob/main/src/test/resources/application.conf#L29 - rest all looks fine :)

vinoocs (changes requested):
Looks good, @Amar - the journey test needs to be updated here https://github.com/hmrc/trader-goods-profiles-ui-journey-tests/blob/main/src/test/resources/application.conf#L29 - rest all looks fine :)

liandra-giacomine (dismissed):

vinoocs (dismissed):

yohan-oc (dismissed):

app/controllers/AdviceStartController.scala
```diff
@@ -38,12 +38,14 @@ class AdviceStartController @Inject() (
...
     with I18nSupport {
 
-  def onPageLoad: Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
-    Ok(view())
+  def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request =>
+      Ok(view(recordId))
   }
 
   //TODO navigate to correct page
```
ChrisCookOC: is todo still valid? if so does it need to be moved to the navigator?

app/controllers/NameController.scala
```diff
@@ -46,26 +46,29 @@ class NameController @Inject() (
...
+          formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, recordId))),
           value =>
             for {
-              updatedAnswers <- Future.fromTry(request.userAnswers.set(NamePage, value))
+              updatedAnswers <- Future.fromTry(request.userAnswers.set(NamePage(recordId), value))
               _              <- sessionRepository.set(updatedAnswers)
-            } yield Redirect(navigator.nextPage(NamePage, mode, updatedAnswers)) //TODO navigate to correct page
+            } yield Redirect(
+              navigator.nextPage(NamePage(recordId), mode, updatedAnswers)
+            ) //TODO navigate to correct page
```
ChrisCookOC: is todo still valid? if so does it need to be moved to the navigator?

app/pages/AdviceStartPage.scala
```diff
@@ -16,4 +16,11 @@
 
 package pages
 
-case object AdviceStartPage extends Page
+import play.api.libs.json.JsPath
+
+case class AdviceStartPage(recordId: String) extends QuestionPage[String] {
```
ChrisCookOC: Don't think this needs to be a QuestionPage? User enters no answer

can't it just be 

case class AdviceStartPage(recordId: String) extends Page

app/config/FrontendAppConfig.scala
```diff
@@ -40,7 +40,7 @@ class FrontendAppConfig @Inject() (configuration: Configuration) {
   val signOutContinueUrl: String = configuration.get[String]("urls.signOutContinue")
 
   private val exitSurveyBaseUrl: String = configuration.get[Service]("microservice.services.feedback-frontend").baseUrl
-  val exitSurveyUrl: String             = s"$exitSurveyBaseUrl/feedback/trader-goods-profiles-frontend"
+  val exitSurveyUrl: String             = s"$exitSurveyBaseUrl/feedback-form/trader-goods-profiles-frontend"
```
liandra-giacomine: Should just be /feedback/<service-name> so what we had before was fine

https://github.com/hmrc/feedback-frontend?tab=readme-ov-file#service-manager

app/controllers/CyaRequestAdviceController.scala
```diff
@@ -45,36 +45,38 @@ class CyaRequestAdviceController @Inject() (
...
 
-  def onSubmit(): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
-    AdviceRequest.build(request.userAnswers, request.eori) match {
-      case Right(model) =>
-        accreditationConnector
-          .submitRequestAccreditation(model)
-          .map(_ => Redirect(routes.AdviceSuccessController.onPageLoad().url))
-      case Left(errors) => Future.successful(logErrorsAndContinue(errors))
-    }
+  def onSubmit(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData).async {
```
liandra-giacomine, line 72: What happens currently when we try to request advice? There's a link on the homepage that might have broken due to this change

app/controllers/NameController.scala
AmarkumarJagtap: removed

app/controllers/AdviceStartController.scala
AmarkumarJagtap: think todo is still valid and it has been on onsubmit def, not sure if we need to move it inside the def and above redirect, as onsubmit has only redirect in it nothing else, any thoughts?

app/pages/AdviceStartPage.scala
AmarkumarJagtap: tried with just page but it has now recordId in path which isn't allowing me to use Page unless we modify Page.scala itself

app/controllers/CyaRequestAdviceController.scala
AmarkumarJagtap: yes had a discussion with Yohan and Beth on this and we feel there shouldn't be request advice on homepage now, rather it should be GoodsRecordsController with default page 1, any thoughts?

app/pages/AdviceStartPage.scala
ChrisCookOC: Should just work as a Page. See e.g. https://github.com/hmrc/trader-goods-profiles-frontend/pull/172/files#diff-c9b5d4c4d6c81750f6d06c5c48617d347aa3ed55fcc65cd3e950e31b9f0480c2

app/controllers/AdviceStartController.scala
AmarkumarJagtap: removed

app/pages/AdviceStartPage.scala
AmarkumarJagtap: now extending with Page

app/controllers/CyaRequestAdviceController.scala
AmarkumarJagtap: removed link from HomePage

ChrisCookOC (dismissed):

ChrisCookOC (dismissed):

app/models/AdviceRequest.scala
```diff
@@ -33,12 +33,12 @@ object AdviceRequest {
...
   implicit lazy val format: OFormat[AdviceRequest] = Json.format
 
-  def build(answers: UserAnswers, eori: String): EitherNec[ValidationError, AdviceRequest] =
+  def build(answers: UserAnswers, eori: String, recordId: String): EitherNec[ValidationError, AdviceRequest] =
     (
       Right(eori),
-      answers.getPageValue(NamePage),
+      answers.getPageValue(NamePage(recordId)),
       Right(eori),
       Right("b0082f50-f13b-416a-8071-3bd95107d44d"),
```
yohan-oc: We can remove the hard coded recordId now and set the recordId parameter here `Right(recordId) `

yohan-oc (dismissed):
LGTM

yohan-oc (approved):

PR #173, opened 2024-07-01
TGP-1458: Update commodity code page url

mycp98:


conf/app.routes
```diff
@@ -53,10 +53,10 @@ POST        /ukims-number/check                              controllers.UkimsNu
 GET         /update/:recordId/categorisation-start           controllers.CategoryGuidanceController.onPageLoad(recordId: String)
 POST        /update/:recordId/categorisation-start           controllers.CategoryGuidanceController.onSubmit(recordId: String)
 
-GET         /create-record/commodity-code                    controllers.CommodityCodeController.onPageLoad(mode: Mode = NormalMode)
-POST        /create-record/commodity-code                    controllers.CommodityCodeController.onSubmit(mode: Mode = NormalMode)
-GET         /create-record/commodity-code/check              controllers.CommodityCodeController.onPageLoad(mode: Mode = CheckMode)
-POST        /create-record/commodity-code/check              controllers.CommodityCodeController.onSubmit(mode: Mode = CheckMode)
+GET         /update-record/commodity-code                    controllers.CommodityCodeController.onPageLoad(mode: Mode = NormalMode)
```
AmarkumarJagtap: @besscerule so is this part of create-record journey or update-record because as part of TGP-1515 it still refers as create-record?

ChrisCookOC (changes requested):
Not sure about this right now. I think it needs to be accessible from the update journey but it also still needs to be in the create journey. So don't think we should be removing the create urls?

ChrisCookOC (dismissed):
Seems we have more issues ignore me

yohan-oc (dismissed):
LGTM

app/controllers/CommodityCodeController.scala
```diff
@@ -52,21 +52,60 @@ class CommodityCodeController @Inject() (
...
+      val onSubmitAction: Call = routes.CommodityCodeController.onSubmitCreate(mode)
+      form
+        .bindFromRequest()
+        .fold(
+          formWithErrors => Future.successful(BadRequest(view(formWithErrors, onSubmitAction))),
+          value =>
+            (for {
+              commodity               <-
+                ottConnector.getCommodityCode(value, request.eori, request.affinityGroup, CreateRecordJourney, None)
+              updatedAnswers          <- Future.fromTry(request.userAnswers.set(CommodityCodePage, value))
```
liandra-giacomine: I think for the update we need a different page value for the session data, CommodityCodePage(recordId) - so that we don't display the value they use in the create journey when they're updating a specific record

conf/app.routes
```diff
@@ -53,10 +53,15 @@ POST        /create-profile/ukims-number/check
...
-GET         /create-record/commodity-code/check                                             controllers.CommodityCodeController.onPageLoad(mode: Mode = CheckMode)
-POST        /create-record/commodity-code/check                                             controllers.CommodityCodeController.onSubmit(mode: Mode = CheckMode)
+GET         /create-record/commodity-code                                                   controllers.CommodityCodeController.onPageLoadCreate(mode: Mode = NormalMode)
+POST        /create-record/commodity-code                                                   controllers.CommodityCodeController.onSubmitCreate(mode: Mode = NormalMode)
+GET         /create-record/commodity-code/check                                             controllers.CommodityCodeController.onPageLoadCreate(mode: Mode = CheckMode)
+POST        /create-record/commodity-code/check                                             controllers.CommodityCodeController.onSubmitCreate(mode: Mode = CheckMode)
+
+GET         /update-record/:recordId/commodity-code          controllers.CommodityCodeController.onPageLoadUpdate(mode: Mode = NormalMode, recordId: String)
```
AmarkumarJagtap: just the minor one to get the indentation right from `controllers.` to be line with above ones

app/navigation/Navigator.scala
```diff
@@ -40,7 +40,7 @@ class Navigator @Inject() () {
     case TraderReferencePage         => _ => routes.UseTraderReferenceController.onPageLoad(NormalMode)
     case UseTraderReferencePage      => navigateFromUseTraderReference
     case GoodsDescriptionPage        => _ => routes.CountryOfOriginController.onPageLoad(NormalMode)
-    case CountryOfOriginPage         => _ => routes.CommodityCodeController.onPageLoad(NormalMode)
+    case CountryOfOriginPage         => _ => routes.CommodityCodeController.onPageLoadCreate(NormalMode)
```
AmarkumarJagtap, line 43: do we need to add one for onPageLoadUpdate method?

mycp98: i've left it for now, i think navigation will be implemented in another ticket?

app/queries/CommodityUpdateQuery.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package queries
+
+import models.Commodity
+import play.api.libs.json.JsPath
+
+object CommodityUpdateQuery extends Gettable[Commodity] with Settable[Commodity] {
```
AmarkumarJagtap: do we need to pass recordId here?

ChrisCookOC: Yes. Can also be a case class

test/pages/CommodityCodeUpdatePageSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+      val result = userAnswers.set(CommodityCodePage, "1234").success.value
+
+      result.isDefined(HasCorrectGoodsPage) mustBe false
+
+    }
+
+    "retains HasCorrectGoodsPage when commodity code has not changed" in {
+
+      val userAnswers = UserAnswers(userAnswersId)
+        .set(CommodityCodePage, "123")
```
AmarkumarJagtap: should this be CommityCodeUpdatePage?

test/pages/CommodityCodeUpdatePageSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+        .success
+        .value
+        .set(HasCorrectGoodsPage, true)
+        .success
+        .value
+
+      val result = userAnswers.set(CommodityCodePage, "123").success.value
```
AmarkumarJagtap: should this be CommityCodeUpdatePage?

test/pages/CommodityCodeUpdatePageSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class CommodityCodeUpdatePageSpec extends AnyFreeSpec with Matchers with TryValues with OptionValues {
+  "clean up" - {
+
+    "removes HasCorrectGoodsPage when commodity code has changed" in {
+
+      val userAnswers = UserAnswers(userAnswersId)
+        .set(CommodityCodePage, "123")
```
AmarkumarJagtap: should this be CommityCodeUpdatePage?

app/queries/CommodityUpdateQuery.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  override def path: JsPath = JsPath \ "commodity"
```
ChrisCookOC: Will want to give it another name here. This is the JSON name in the User Answers and its shared with the original CommodityQuery. So if you use "commodity" here it will overwrite the CommodityQuery data and they will start reading each others data etc

When you add the recordId you'll want to stick that in the path anyway, so would probably avoid overwrite, but I'd give it another name for niceness sake anyway :)

test/pages/CommodityCodeUpdatePageSpec.scala
mycp98: ive updated this now, thanks 😄

app/queries/CommodityUpdateQuery.scala
AmarkumarJagtap: commodityUpdate or may be commdityCodeUpdate ?

app/controllers/CommodityCodeController.scala
```diff
@@ -52,21 +52,37 @@ class CommodityCodeController @Inject() (
 
   private val form = formProvider()
 
-  def onPageLoad(mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
-    val preparedForm = request.userAnswers.get(CommodityCodePage) match {
-      case None        => form
-      case Some(value) => form.fill(value)
-    }
+  def onPageLoadCreate(mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
```
AmarkumarJagtap, line 55: looks good :), just this one needs scala formatting for CommodityCodeController.scala

AmarkumarJagtap (dismissed):
Looks good to me, the only change is required on CommodityCodeUpdatePage.scala to have HasCorrectGoodsPage to have recordId being passed but that should be covered as part of TGP-1459 :)

liandra-giacomine (dismissed):

ChrisCookOC (dismissed):

AmarkumarJagtap (approved):

PR #176, opened 2024-07-02
TGP-1182: updated commodity code result

vinoocs:
- updated the commodity object to receive multiple descriptions (maximum 3)

- **the functionality for  /update/:recordId/longer-commodity-code-result will be a separate ticket as discussed with Amaka so please ignore that in this**

Page:
image

### DIFF
--- a/app/views/HasCorrectGoodsView.scala.html
+++ b/app/views/HasCorrectGoodsView.scala.html
@@ -35,7 +35,7 @@
 
     <ul class="govuk-body">
         @commodity.descriptions.zipWithIndex.map { case (description, index) =>
-            <li>@messages(description)</li>
+            <li>@messages(description.toLowerCase)</li>
         }
     </ul>
 
--- a/app/views/HasCorrectGoodsView.scala.html
+++ b/app/views/HasCorrectGoodsView.scala.html
@@ -35,7 +35,7 @@
 
     <ul class="govuk-body">
         @commodity.descriptions.zipWithIndex.map { case (description, index) =>
-            <li>@messages(description.toLowerCase)</li>
+            <li>@description.toLowerCase</li>
         }
     </ul>
 
--- a/app/views/HasCorrectGoodsView.scala.html
+++ b/app/views/HasCorrectGoodsView.scala.html
@@ -34,9 +34,9 @@
     <p class="govuk-body">@messages("hasCorrectGoods.p1")</p>
 
     <ul class="govuk-body">
-    @commodity.descriptions.zipWithIndex.map { case (description, index) =>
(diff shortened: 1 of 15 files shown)


app/views/HasCorrectGoodsView.scala.html
```diff
@@ -30,10 +30,18 @@
 
     <h1 class="govuk-heading-l">@messages("hasCorrectGoods.heading", commodity.commodityCode)</h1>
 
-    <p class="govuk-body">@messages("hasCorrectGoods.p1", commodity.description)</p>
+
+    <p class="govuk-body">@messages("hasCorrectGoods.p1")</p>
+
+    <ul class="govuk-body">
+    @commodity.descriptions.zipWithIndex.map { case (description, index) =>
```
ChrisCookOC: can do some indenting here to make more readable

test/factories/AuditEventFactorySpec.scala
```diff
@@ -243,7 +244,7 @@ class AuditEventFactorySpec extends SpecBase {
 
           val effectiveFrom = Instant.now
           val effectiveTo   = effectiveFrom.plusSeconds(99)
-          val commodity     = Commodity("030821", "Sea urchins", effectiveFrom, Some(effectiveTo))
+          val commodity     = Commodity("030821", List("Sea urchins"), effectiveFrom, Some(effectiveTo))
```
ChrisCookOC: If you really wanted to (but its not important) you could have multiple descriptions in the list and then you are testing that the audit event gets the head like you want.

ChrisCookOC (commented):
Looks good, minor thoughts added

The Json stuff does make me a bit 🫠, though I know its very hard to make nice looking code when Json is involved here 😄

app/views/HasCorrectGoodsView.scala.html
```diff
@@ -30,10 +30,18 @@
 
     <h1 class="govuk-heading-l">@messages("hasCorrectGoods.heading", commodity.commodityCode)</h1>
 
-    <p class="govuk-body">@messages("hasCorrectGoods.p1", commodity.description)</p>
+
+    <p class="govuk-body">@messages("hasCorrectGoods.p1")</p>
+
+    <ul class="govuk-body">
+        @commodity.descriptions.zipWithIndex.map { case (description, index) =>
+            <li>@messages(description.toLowerCase)</li>
```
ChrisCookOC: if this is coming from OTT why does it need to go through messages?

ChrisCookOC (approved):

liandra-giacomine (approved):

PR #177, opened 2024-07-02
TGP-1423: investigate loading pages

besscerule:


app/controllers/GoodsRecordsController.scala
```diff
@@ -48,36 +47,48 @@ class GoodsRecordsController @Inject() (
     extends FrontendBaseController
     with I18nSupport {
 
-  private val form = formProvider()
+  private val form     = formProvider()
+  private val pageSize = 10
```
ChrisCookOC, line 51: Could maybe make this a app config variable and then it might be easier to test pagination etc in future. Not big deal though.

app/controllers/PreviousMovementRecordsController.scala
```diff
@@ -18,50 +18,38 @@ package controllers
...
-        Ok(view())
-
-      case Some(_) =>
-        Redirect(routes.GoodsRecordsController.onPageLoadNoRecords())
+    goodsRecordConnector.getRecordsCount(request.eori).flatMap {
+      case 0 => Future.successful(Redirect(routes.GoodsRecordsController.onPageLoadNoRecords()))
+      case _ =>
+        goodsRecordConnector.doRecordsExist(request.eori).map {
+          case false => Ok(view())
+          case true  => Redirect(routes.GoodsRecordsController.onPageLoad(1))
```
ChrisCookOC: Perhaps make the 1 a "firstGoodsPage" variable?

app/models/GoodsRecordsPagination.scala
```diff
@@ -32,47 +31,67 @@ case class GoodsRecordsPagination(
 
 object GoodsRecordsPagination {
   implicit val format: OFormat[GoodsRecordsPagination] = Json.format[GoodsRecordsPagination]
+  private val defaultRecord                            = 0
 
-  private def getPageSize(totalRecords: Int, recordsSize: Int, totalPages: Int): Int =
-    (totalRecords - recordsSize) / (totalPages - 1)
+  def getFirstRecord(pagination: GoodsRecordsPagination, pageSize: Int): Int =
```
ChrisCookOC: Maybe a slightly different name that makes it clear its like a goods record index and not the record itself? Seeing the calls for this I was expecting it to return a record, not an int

app/models/GoodsRecordsPagination.scala
```diff
@@ -32,47 +31,67 @@ case class GoodsRecordsPagination(
...
-      items = Some((0 until pagination.totalPages).map { index =>
-        PaginationItem(
-          number = Some((index + 1).toString()),
-          current = Some(pagination.currentPage == (index + 1)),
-          href = routes.GoodsRecordsController.onPageLoad(index + 1).url
-        )
-      }),
-      previous = if (pagination.currentPage == 1) {
-        None
+      val start = if (currentPage <= 2) {
```
ChrisCookOC, line 59: not sure if doing like Max(1, currentPage-2) might be more comprehensible. This is good though, just took me some playing to figure out what everything meant and why -2 etc was floating around. Impressed by how much hassle it is just to get that little page number thing at the bottom of the list!

it/test/connectors/GoodsRecordConnectorSpec.scala
```diff
@@ -473,79 +478,149 @@ class GoodsRecordConnectorSpec
...
-          .willReturn(ok().withBody(getMultipleRecordResponseData.toString()))
+        get(urlEqualTo(getGoodsRecordsCountUrl))
+          .withHeader(xClientIdName, equalTo(xClientId))
+          .willReturn(ok().withBody("3"))
+      )
+
+      connector.getRecordsCount(testEori).futureValue mustBe 3
+    }
+
+    "must retunr 0 if there are no records" in {
```
ChrisCookOC: typo in return

test/models/GoodsRecordsPaginationSpec.scala
```diff
@@ -0,0 +1,942 @@
...
+
+class GoodsRecordsPaginationSpec extends AnyFreeSpec with Matchers with TryValues with OptionValues {
+  private val pageSize = 10
+
+  ".getPagination" - {
+
+    "when currentPage 0" in {
+      val currentPage = 0
+      val totalPages  = 4
+      val pagination  = Pagination(
```
ChrisCookOC: These tests might all be simpler/clearer if we name it 
val expectedPagination

and then the result.blah mustEqual paginatioin.blah
could just be result mustEqual expectedPagination

ChrisCookOC:
Looking good :)

app/controllers/PreviousMovementRecordsController.scala
vinoocs: wondering if the design has changed for this? My understanding is that it should always go through previous movement records page if records exist and redirect happens on load records button.

app/controllers/GoodsRecordsController.scala
besscerule: it doesn't change between environments so I'll keep it as it is for the moment

app/controllers/PreviousMovementRecordsController.scala
besscerule: so if they have already been stored it should redirect and not see this page at all

test/models/GoodsRecordsPaginationSpec.scala
besscerule: done

it/test/connectors/GoodsRecordConnectorSpec.scala
besscerule: done

app/models/GoodsRecordsPagination.scala
besscerule: i'll just leave it as it is

app/models/GoodsRecordsPagination.scala
besscerule: Fixed

app/controllers/PreviousMovementRecordsController.scala
```diff
@@ -31,36 +29,28 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    for {
-      getGoodsRecordResponse  <- getGoodsRecordConnector.getAllRecords(request.eori)
-      updatedAnswersWithQuery <- Future.fromTry(request.userAnswers.set(GetGoodsRecordsQuery, getGoodsRecordResponse))
-      updatedAnswersWithFlag  <-
-        Future.fromTry(updatedAnswersWithQuery.set(PreviousMovementRecordsPage, "hasLoadedRecords"))
-      _                       <- sessionRepository.set(updatedAnswersWithFlag)
-
-    } yield Redirect(routes.GoodsRecordsController.onPageLoad(1))
-
+    goodsRecordConnector.storeAllRecords(request.eori).map(_ => Redirect(routes.GoodsRecordsController.onPageLoad(1)))
```
ChrisCookOC: This could use the firstPage constant

ChrisCookOC (dismissed):

ChrisCookOC (approved):

PR #182, opened 2024-07-03
Tgp 1541/navigation journey2

ChrisCookOC:
Hoping the code speaks for itself but I appreciate my brain has been in this for 3 days and so some explanations and pretty pictures might help

**Flow**
image

**Why is the OnSubmit in the HasCorrectGoodsController doing all that**
If the user selects Yes to their Longer Commodity Code, we need to a) see if the goods need recategorising and b) ensure that when we next categorise the goods we have the right assessment details to show them

So we grab the Shorter Commodity Code's Assessment Details (Line 123), and we go off to OTT and get the Longer Commodity Code's Assessment Details (Line 127 - 132). This will also update the RecordCategorisations map in the user answers to have the Longer Commodity Code details in, satisfying (b)

Having got the Shorter and Longer Assessment Details, we can compare them to see if we need to recategorise (Line 137), satisfying (a).
image

If we do need to recategorise, then we need to remove the old Assessment answers from the User Answers (Line 140). This is because a) we don't want the assessment answers pre-populating and b) the Check Your Answers will be unhappy with old invalid data floating around

Having done all this, we have a nice up to date User Answers and can go onto the Navigator to tell us what to do next.

**Supplementary unit**
This story only applies to cases with no supplementary unit involved. That comes in part 3.


app/controllers/HasCorrectGoodsController.scala
```diff
@@ -53,33 +58,37 @@ class HasCorrectGoodsController @Inject() (
...
   }
   // TODO - this is still not functional, it is just to create the url. Implement this properly
   def onPageLoadLongerCommodityCode(mode: Mode, recordId: String): Action[AnyContent] =
     (identify andThen getData andThen requireData) { implicit request =>
-      val preparedForm = request.userAnswers.get(HasCorrectGoodsPage) match {
+      val preparedForm = request.userAnswers.get(HasCorrectGoodsLongerCommodityCodePage(recordId)) match {
         case None        => form
         case Some(value) => form.fill(value)
       }
       //TODO - use the UpdateCommdityQuery (with recordId) here when available
```
liandra-giacomine: Do we still need this TODO?

app/controllers/HasCorrectGoodsController.scala
```diff
@@ -92,20 +101,92 @@ class HasCorrectGoodsController @Inject() (
...
+              case Some(commodity) => Future.successful(BadRequest(view(formWithErrors, commodity, submitAction)))
               case None            => Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url))
             },
           value =>
-            for {
-              updatedAnswers <- Future.fromTry(request.userAnswers.set(HasCorrectGoodsPage, value))
-              _              <- sessionRepository.set(updatedAnswers)
-            } yield Redirect(navigator.nextPage(HasCorrectGoodsPage, mode, updatedAnswers))
+            if (value) {
+              for {
```
liandra-giacomine: wonder if it's worth putting the logic in this for comprehension into a function

liandra-giacomine (commented):
Looks good, just makes me think that it may be helpful to have something like a categorisation service or something but would only really make sense if we share functionality across diff controllers

Ewan-James-Donovan (commented):
I'm happy with this so far, it seems to be playing well with my check-your-answers stuff, too. I'll have a proper review today at some point, I still want time to parse what's going on here and there's also a couple conflicts.

Props to you, tackled this like a champ!

app/controllers/HasCorrectGoodsController.scala
ChrisCookOC: Nope, removed :)

app/controllers/HasCorrectGoodsController.scala
ChrisCookOC: Certainly neatens things up . Done

app/controllers/HasCorrectGoodsController.scala
```diff
@@ -148,4 +163,72 @@ class HasCorrectGoodsController @Inject() (
             } yield Redirect(navigator.nextPage(HasCorrectGoodsCommodityCodeUpdatePage(recordId), mode, updatedAnswers))
         )
     }
+
+  private def processNewCommodityCode(mode: Mode, recordId: String, request: DataRequest[AnyContent], value: Boolean)(
+    implicit hc: HeaderCarrier
+  ) =
+    for {
+      // Save the user's answer
```
Ewan-James-Donovan, line 171: I personally love comments, but should we have them? Up to you

Ewan-James-Donovan (approved):

app/controllers/HasCorrectGoodsController.scala
ChrisCookOC: I'm hoping they are useful and demystify the code here, so I'm happy if no one is really unhappy 🙂

mycp98 (approved):

app/controllers/HasCorrectGoodsController.scala
mycp98: i like the comments 😄

PR #188, opened 2024-07-03
TGP-1455: Update record check your answers

besscerule:


app/navigation/Navigator.scala
```diff
@@ -28,20 +29,19 @@ import scala.util.Try
...
     case ProfileSetupPage            => _ => routes.UkimsNumberController.onPageLoad(NormalMode)
     case UkimsNumberPage             => _ => routes.HasNirmsController.onPageLoad(NormalMode)
     case HasNirmsPage                => navigateFromHasNirms
     case NirmsNumberPage             => _ => routes.HasNiphlController.onPageLoad(NormalMode)
     case HasNiphlPage                => navigateFromHasNiphl
     case NiphlNumberPage             => _ => routes.CyaCreateProfileController.onPageLoad
-    case CreateRecordStartPage       => _ => routes.TraderReferenceController.onPageLoad(NormalMode)
-    case TraderReferencePage         => _ => routes.UseTraderReferenceController.onPageLoad(NormalMode)
+    case CreateRecordStartPage       => _ => routes.TraderReferenceController.onPageLoad(NormalMode, newRecordId)
+    case p: TraderReferencePage      => _ => routes.UseTraderReferenceController.onPageLoad(NormalMode)
```
liandra-giacomine: what's this `p:` thing about? :D

liandra-giacomine: Ah I see it's used in some cases but not in these

conf/app.routes
```diff
@@ -53,15 +53,18 @@ POST        /create-profile/ukims-number/check
 GET         /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onPageLoad(recordId: String)
 POST        /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onSubmit(recordId: String)
 
-GET         /create-record/commodity-code                                                   controllers.CommodityCodeController.onPageLoad(mode: Mode = NormalMode)
-POST        /create-record/commodity-code                                                   controllers.CommodityCodeController.onSubmit(mode: Mode = NormalMode)
-GET         /create-record/commodity-code/check                                             controllers.CommodityCodeController.onPageLoad(mode: Mode = CheckMode)
-POST        /create-record/commodity-code/check                                             controllers.CommodityCodeController.onSubmit(mode: Mode = CheckMode)
+GET         /create-record/commodity-code                                                   controllers.CommodityCodeController.onPageLoad(mode: Mode = NormalMode, recordId: String = "new-record")
```
liandra-giacomine: why are we defaulting the string here?

besscerule: because it needs to be explicit, scala doesn't let you default it in the controller onPageLoad unfortunately

AmarkumarJagtap: just thinking do we need to even pass recordId as parameter in here, and may be create separate onLoad and OnSubmit for create and update? something like this https://github.com/hmrc/trader-goods-profiles-frontend/blob/main/app/controllers/HasCorrectGoodsController.scala

ChrisCookOC: Think I'd lean towards having a version that doesn't take in RecordId and one that does, rather than hardcoding a dummy one

liandra-giacomine: Yeah I don't see why we are passing a record Id to the create journey?

liandra-giacomine: I see what Beth has done here, she's reusing the page components - which is okay but instead of a default string, it should be optional.

liandra-giacomine (dismissed):
Refactored in another PR as we don't want to use query params here

app/navigation/Navigator.scala
```diff
@@ -38,11 +37,20 @@ class Navigator @Inject() () {
...
-    case GoodsDescriptionPage                      => _ => routes.CountryOfOriginController.onPageLoad(NormalMode)
+    case GoodsDescriptionPage                      => _ => routes.CountryOfOriginController.onPageLoadCreate(NormalMode)
+    case p: GoodsDescriptionUpdatePage             => _ => routes.CyaUpdateRecordController.onPageLoadGoodsDescription(p.recordId)
     case CountryOfOriginPage                       => _ => routes.CommodityCodeController.onPageLoadCreate(NormalMode)
+    case p: CountryOfOriginUpdatePage              => _ => routes.CyaUpdateRecordController.onPageLoadCountryOfOrigin(p.recordId)
     case CommodityCodePage                         => _ => routes.HasCorrectGoodsController.onPageLoadCreate(NormalMode)
-    case HasCorrectGoodsPage                       => navigateFromHasCorrectGoods
+    case p: CommodityCodeUpdatePage                =>
+      _ => {
+        println("going to has goods")
```
ChrisCookOC: println :)

ChrisCookOC (approved):

PR #201, opened 2024-07-08
TGP-1391:cleanse data on submit

vinoocs:
- Refactored as per comments
- Also fixes https://jira.tools.tax.service.gov.uk/browse/TGP-1900 

**Note :** Not done updateJourney in this ticket as it needs a small clarification from Amaka, also it's not scope for this ticket so can wait. But this fix needs to go for a bug raised.


app/controllers/CyaRequestAdviceController.scala
```diff
@@ -56,27 +59,33 @@ class CyaRequestAdviceController @Inject() (
...
+    errors: data.NonEmptyChain[ValidationError],
+    recordId: String,
+    request: DataRequest[AnyContent]
+  ): Result = {
     val errorMessages = errors.toChain.toList.map(_.message).mkString(", ")
 
     val continueUrl = RedirectUrl(routes.AdviceStartController.onPageLoad(recordId).url)
 
     logger.warn(s"Unable to create Request Advice.  Missing pages: $errorMessages")
+    dataCleansingService.deleteMongoData(request.userAnswers.id)
```
Ewan-James-Donovan: We don't seem to be awaiting or handling the Future[Boolean] returned from dataCleansingService -> Could this cause issues or not?

app/controllers/CyaCategorisationController.scala
```diff
@@ -105,6 +107,7 @@ class CyaCategorisationController @Inject() (
             model.categoryAssessmentsWithExemptions,
             model.category
           )
+          dataCleansingService.deleteMongoData(request.userAnswers.id)
```
ChrisCookOC: Should this happen after the goodsRecordConnector call? Just wondering if we want the data cleared if that fails?

app/controllers/CyaCreateProfileController.scala
```diff
@@ -62,28 +65,31 @@ class CyaCreateProfileController @Inject() (
...
-        case Left(errors) => logErrorsAndContinue(errors)
+        case Left(errors) => logErrorsAndContinue(errors, request)
       }
   }
 
   def onSubmit(): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
     TraderProfile.build(request.userAnswers, request.eori) match {
       case Right(model) =>
         auditService.auditProfileSetUp(model, request.affinityGroup)
+        dataCleansingService.deleteMongoData(request.userAnswers.id)
```
ChrisCookOC: same query as above re: connector

app/repositories/SessionRepository.scala
```diff
@@ -89,4 +90,16 @@ class SessionRepository @Inject() (
       .deleteOne(byId(id))
       .toFuture()
       .map(_ => true)
+
+  def clearData(id: String): Future[Boolean] =
+    collection
+      .updateOne(
+        filter = byId(id),
+        update = Updates.combine(
+          Updates.set("data", Document()),
```
ChrisCookOC: Not sure this is nuanced enough for those bits of data that depend on record id - we decided to store it all by record id because we thought multiple people might use it or different items will be in progress, but this just wipes it all whenever anything finishes?

ChrisCookOC (commented):
(not looked at tests)

app/controllers/CyaCategorisationController.scala
liandra-giacomine: We would always want to clear it if it's a validation failure and in the scenario where it succeeds. I can see a scenario where it would be useful to keep it being if we get a 500 if B&T is down but I think for now this is okay/not over engineering/see if it comes up in the feedback

liandra-giacomine: That's a good point, let's move this inside the map of the connector calls

app/controllers/CyaRequestAdviceController.scala
liandra-giacomine: If this call fails, we would take the user to an error page instead of journey recovery but the session data would still be there...I think it's ok for this to be a fire and forget but haven't thought about all the pros/cons

app/repositories/SessionRepository.scala
liandra-giacomine: Agreed, this needs to be specific to the journey i.e. delete all pages that get populated in the create profile journey if we're in the create profile CYA. We probably need specific clearData functions for each journey, I suppose instead of update you can just do a delete

liandra-giacomine: you could have an enum which holds pages for each journey and match on that in the clearData, loop through the pages held in a list and delete each


Looks like we already have an object for that which could be extended:

```
sealed trait Journey

case object CreateRecordJourney extends Journey {
 val pages = Seq(CommodityCodePage, ...)
  override def toString: String = "CreateRecord"
}
case object UpdateRecordJourney extends Journey {
   deg pages(recordId) = Seq(CommodityCodePage(recordId, ...)
  override def toString: String = "UpdateRecord"
}



def clearData(journey: Journey) = journey match {
case UpdateRecordJourney => UpdateRecordJourney.pages.map(page => collection.delete(page))
```

ChrisCookOC: I like the idea of extending these 🙂

liandra-giacomine (dismissed):
Approving this, we can look into refactoring later if we want to

Ewan-James-Donovan (dismissed):

app/controllers/CyaRequestAdviceController.scala
vinoocs: yeah just fire and forget same like audit service!

app/controllers/CyaCategorisationController.scala
vinoocs: moved it inside connector calls

app/repositories/SessionRepository.scala
vinoocs: refactored to be very specific to the journey!

liandra-giacomine (dismissed):

liandra-giacomine (approved):

PR #208, opened 2024-07-09
Tgp 1457/implement navigation for change record warning page country of origin

omotrium:


app/controllers/SingleRecordController.scala
```diff
@@ -42,10 +42,10 @@ class SingleRecordController @Inject() (
...
       val detailsList = SummaryListViewModel(
         rows = Seq(
-          TraderReferenceSummary.row(record.traderRef, recordId, NormalMode),
-          GoodsDescriptionSummary.row(record.goodsDescription, recordId, NormalMode),
-          CountryOfOriginSummary.row(record.countryOfOrigin, recordId, NormalMode),
-          CommodityCodeSummary.row(record.commodityCode, recordId, NormalMode),
+          TraderReferenceSummary.row(record.traderRef, recordId, CheckMode),
+          GoodsDescriptionSummary.row(record.goodsDescription, recordId, CheckMode),
+          CountryOfOriginSummary.row(record.countryOfOrigin, recordId, CheckMode),
+          CommodityCodeSummary.row(record.commodityCode, recordId, CheckMode),
```
yohan-oc: This should be `NormalMode`

app/navigation/Navigator.scala
```diff
@@ -197,6 +197,7 @@ class Navigator @Inject() () {
     case p: EmailPage                              => _ => routes.CyaRequestAdviceController.onPageLoad(p.recordId)
     case p: AssessmentPage                         => navigateFromAssessmentCheck(p)
     case p: HasSupplementaryUnitPage               => navigateFromHasSupplementaryUnitCheck(p.recordId)
+    case p: HasCountryOfOriginChangePage           => answers => navigateFromHasCountryOfOriginChangeCheck(answers, p.recordId)
```
yohan-oc: Missing unit test for this

besscerule: This should also be in the normal mode section not the check mode

app/navigation/Navigator.scala
```diff
@@ -219,6 +220,20 @@ class Navigator @Inject() () {
...
 
+  private def navigateFromHasCountryOfOriginChangeCheck(answers: UserAnswers, recordId: String): Call =
+    answers
+      .get(HasCountryOfOriginChangePage(recordId))
+      .map {
+        case true  =>
+          if (answers.isDefined(CountryOfOriginUpdatePage(recordId))) {
+            routes.CyaUpdateRecordController.onPageLoadCountryOfOrigin(recordId)
+          } else {
+            routes.CountryOfOriginController.onPageLoadUpdate(CheckMode, recordId)
```
besscerule: should be normal mode

app/viewmodels/checkAnswers/CountryOfOriginSummary.scala
```diff
@@ -46,7 +46,7 @@ object CountryOfOriginSummary {
       key = "countryOfOrigin.checkYourAnswersLabel",
       value = ValueViewModel(HtmlFormat.escape(value).toString),
       actions = Seq(
-        ActionItemViewModel("site.change", routes.CountryOfOriginController.onPageLoadUpdate(mode, recordId).url)
+        ActionItemViewModel("site.change", routes.HasCountryOfOriginChangeController.onPageLoad(mode, recordId).url)
```
besscerule, line 49: i am not sure about this - because this means that when the user clicks on the change link when they are checking the answers it goes right back to the warning page when i think i should go to the country of origin page -> you might need to check with amaka on this

besscerule (changes requested):
Also the user needs to be navigated back to the goods page after they have saved their new answer - currently it goes to the home page

app/viewmodels/checkAnswers/CountryOfOriginSummary.scala
omotrium: the right behavior is for it to go back to the country of origin page. Amaka confirmed it

PR #214, opened 2024-07-10
TGP-373: Create Profile Details Page

yohan-oc:

PR #216, opened 2024-07-10
Tgp 474/create remove nirms/niphl page

omotrium:


app/viewmodels/checkAnswers/HasNiphlsChangeSummary.scala
```diff
@@ -0,0 +1,42 @@
+/*
```
besscerule: we don't need this page as it is never seen in a CYA page

app/viewmodels/checkAnswers/HasNirmsChangeSummary.scala
```diff
@@ -0,0 +1,42 @@
+/*
```
besscerule: see above

conf/app.routes
```diff
@@ -209,3 +209,13 @@ GET         /update-record/:recordId/check-your-answers/trader-reference
 POST        /update-record/:recordId/check-your-answers/trader-reference                    controllers.CyaUpdateRecordController.onSubmitTraderReference(recordId: String)
 GET         /update-record/:recordId/check-your-answers/commodity-code                      controllers.CyaUpdateRecordController.onPageLoadCommodityCode(recordId: String)
 POST        /update-record/:recordId/check-your-answers/commodity-code                      controllers.CyaUpdateRecordController.onSubmitCommodityCode(recordId: String)
+
+GET        /update-profile/removing-nirms-number                                            controllers.HasNirmsChangeController.onPageLoad(mode: Mode = NormalMode)
+POST       /update-profile/removing-nirms-number                                            controllers.HasNirmsChangeController.onSubmit(mode: Mode = NormalMode)
+GET        /changeHasNirmsChange                                                            controllers.HasNirmsChangeController.onPageLoad(mode: Mode = CheckMode)
```
besscerule: in both cases we don't need the CheckMode version of these routes as we never view them from a CYA perspective we always go straight through the journey, so I would remove the CheckMode routes and then also remove the 'mode' parameters for both of the controllers as they are not needed

conf/messages.en
```diff
@@ -451,3 +451,28 @@ hasGoodsDescriptionChange.p1 = Changing the goods description on this goods reco
...
+hasNirmsChange.no = No, remove my NIRMS details
+hasNirmsChange.warning = Warning
+hasNirmsChange.warningText = If you move goods under SPIMM and they are not eligible to, you could be liable for any incorrectly declared goods.
+
+hasNiphlsChange.title = hasNiphlsChange
+hasNiphlsChange.h1 = Removing NIPHL details
+hasNiphlsChange.p1 = If you remove your NIPHL details, the categorisation of goods you move under the Simplified Process for Internal Market Movements (SPIMM) could change.
+hasNiphlsChange.p2 = You will need to re-categorise those records to see if you can still move them under SPIMM.
+hasNiphlsChange.h2 = Are you sure you want to remove your NIPHL details?
+hasNiphlsChange.checkYourAnswersLabel = hasNiphlsChange
```
besscerule: this line isn't needed

conf/messages.en
```diff
@@ -451,3 +451,28 @@ hasGoodsDescriptionChange.p1 = Changing the goods description on this goods reco
...
+hasNiphlsChange.error.required = Confirm if you want to remove your NIPHL
+hasNiphlsChange.change.hidden = HasNiphlsChange
```
besscerule: this line isn't needed

conf/messages.en
```diff
@@ -451,3 +451,28 @@ hasGoodsDescriptionChange.p1 = Changing the goods description on this goods reco
 hasGoodsDescriptionChange.h2 = Do you want to change the goods description?
 hasGoodsDescriptionChange.error.required = Select if you want to change the goods description
 
+
+hasNirmsChange.title = hasNirmsChange
+hasNirmsChange.h1 = Removing NIRMS details
+hasNirmsChange.p1 = If you remove your NIRMS details, the categorisation of goods you move under the Simplified Process for Internal Market Movements (SPIMM) could change.
+hasNirmsChange.p2 = You will need to re-categorise those records to see if you can still move them under SPIMM.
+hasNirmsChange.h2 = Are you sure you want to remove your NIRMS details?
+hasNirmsChange.checkYourAnswersLabel = hasNirmsChange
```
besscerule: this line isn't needed

besscerule (changes requested):
Virtually perfect :)

At some point we will need to update the data store with the new data - i think this will probably be on Submit of your two controllers if the user answers Yes - however I think that is part of the validate or navigation ticket so it doesn't need to be included in this ticket

app/views/HasNiphlsChangeView.scala.html
```diff
@@ -0,0 +1,63 @@
...
+@(form: Form[_], mode: Mode)(implicit request: Request[_], messages: Messages)
+
+@layout(pageTitle = title(form, messages("hasNiphlsChange.title"))) {
+
+        <h1 class="govuk-heading-l">@messages("hasNiphlsChange.h1")</h1>
+
+            <p class="govuk-body">@messages("hasNiphlsChange.p1")</p>
+            <p class="govuk-body">@messages("hasNiphlsChange.p2")</p>
+
+        <div class="govuk-warning-text">
```
vinoocs: can use the warning component as in the other PR https://github.com/hmrc/trader-goods-profiles-frontend/pull/214/files

besscerule (approved):

vinoocs (approved):

PR #223, opened 2024-07-11
TGP-1600 - Niphls on category assessments

ChrisCookOC:
image

All categorisation now must start by loading the Category Guidance, even if it is not actually displayed. This is including Recategorisation (for recategorisation it won't display the page, but it needs the logic to work out if we are doing NIPHLs etc)
-> As part of that change the page now takes a mode, but I never expect the CheckMode version of the page to be displayed to the user. Again its there just because recategorisation needs to route through it.

The RecategorisingQuery is going to be used more in Marcy's upcoming PR because we need different behaviour for loading assessment pages in recategorisation mode


conf/app.routes
```diff
@@ -65,8 +65,10 @@ POST        /create-profile/ukims-number/check
 GET         /update-profile/ukims-number                                                    controllers.UkimsNumberController.onPageLoadUpdate
 POST        /update-profile/ukims-number                                                    controllers.UkimsNumberController.onSubmitUpdate
 
-GET         /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onPageLoad(recordId: String)
-POST        /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onSubmit(recordId: String)
+GET         /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onPageLoad(mode: Mode = NormalMode, recordId: String)
+POST        /update-record/:recordId/categorisation/start                                   controllers.CategoryGuidanceController.onSubmit(mode: Mode = NormalMode, recordId: String)
+GET         /update-record/:recordId/categorisation/start/check                                   controllers.CategoryGuidanceController.onPageLoad(mode: Mode = CheckMode, recordId: String)
```
ChrisCookOC, line 70: I never expect the CheckMode version of the page to be displayed to the user. Again its there just because recategorisation needs to route through it.

app/controllers/CategoryGuidanceController.scala
```diff
@@ -46,20 +47,22 @@ class CategoryGuidanceController @Inject() (
...
-    implicit request =>
+  def onPageLoad(mode: Mode, recordId: String): Action[AnyContent] =
+    (identify andThen getData andThen requireData).async { implicit request =>
       categorisationService
         .requireCategorisation(request, recordId)
         .flatMap { userAnswers =>
           val recordCategorisations = userAnswers.get(RecordCategorisationsQuery)
           val categorisationInfo    = recordCategorisations.flatMap(_.records.get(recordId))
           val scenario              = categorisationInfo.map(Scenario.getRedirectScenarios)
+          val areWeRecategorising   = userAnswers.get(RecategorisingQuery(recordId)).getOrElse(false)
```
liandra-giacomine, line 66: should we be defaulting here? is it not an invalid scenario if we don't have that query

app/models/CategoryRecord.scala
```diff
@@ -51,16 +50,25 @@ object CategoryRecord {
...
+  def buildForNiphls(eori: String, recordId: String, traderProfile: TraderProfile): CategoryRecord = {
+    val category = if (traderProfile.niphlNumber.isDefined) {
+      Category2
+    } else {
+      Category1
+    }
+
+    val categoryAsNumber = if (category == Category1) CATEGORY_1 else CATEGORY_2
+
+    CategoryRecord(eori, recordId, categoryAsNumber, 1)
```
liandra-giacomine: do we want to return the categoryAsNumber here instead of 1?

app/models/ott/CategorisationInfo.scala
```diff
@@ -35,6 +35,11 @@ final case class CategorisationInfo(
     case None                  =>
       false
   }
+
+  def isNiphls: Boolean = categoryAssessments.exists(_.isNiphlsAnswer) &&
```
liandra-giacomine, line 39: Can we add a val to the case class -  same as you've done with isEmptyCat2Assessment

```
val isCategory2 = category == CATEGORY_2

categoryAssessments.count(_.isCategory2)
```

app/models/ott/CategoryAssessment.scala
```diff
@@ -34,6 +36,13 @@ final case class CategoryAssessment(
   def getExemptionListItems: Seq[String] = exemptions.map { exemption =>
     exemption.code + " - " + exemption.description
   }
+
+  def isNiphlsAnswer: Boolean =
+    category == 1 && exemptions.exists(ex =>
```
liandra-giacomine: Can we replace 1s and 2s with the constant CATEGORY1 etc.

app/navigation/Navigator.scala
```diff
@@ -382,38 +389,42 @@ class Navigator @Inject() () {
...
-        if (assessmentPage.index + 1 < assessmentCount) {
-          if (answers.isDefined(AssessmentPage(recordId, assessmentPage.index + 1))) {
-            routes.CyaCategorisationController.onPageLoad(recordId)
+      for {
+        recordQuery      <- answers.get(RecordCategorisationsQuery)
+        record           <- recordQuery.records.get(recordId)
+        assessmentAnswer <- answers.get(assessmentPage)
+      } yield assessmentAnswer match {
+        case AssessmentAnswer.Exemption(_) =>
+          val assessmentCount = Try {
```
liandra-giacomine, line 401: do we want to default to 0 here

app/services/CategorisationService.scala
```diff
@@ -39,122 +39,87 @@ class CategorisationService @Inject() (
...
-            Future.fromTry(cleanupOldAssessmentAnswers(updatedAnswers, recordId))
-          _                                <- sessionRepository.set(updatedAnswersCleanUpAssessments)
-        } yield updatedAnswersCleanUpAssessments
-    }
-  }
+  ): Future[UserAnswers] =
+    (for {
+      getGoodsRecordResponse <- goodsRecordsConnector.getRecord(eori = request.eori, recordId = recordId)
+      recordCategorisations   =
+        request.userAnswers.get(RecordCategorisationsQuery).getOrElse(RecordCategorisations(Map.empty))
```
liandra-giacomine: defaulting here and below, defeats the point of having options

app/services/CategorisationService.scala
```diff
@@ -39,122 +39,87 @@ class CategorisationService @Inject() (
...
-                                  case _                        => Future.failed(new RuntimeException("Could not build categorisation info"))
-                                }
-      updatedAnswers         <-
+      goodsNomenclature       <- ottConnector.getCategorisationInfo(
+                                   commodityCode,
+                                   request.eori,
+                                   request.affinityGroup,
+                                   Some(recordId),
+                                   countryOfOrigin,
+                                   LocalDate.now() //TODO where does DateOfTrade come from??
```
liandra-giacomine: current date time makes sense here I believe, commodity data can change and I believe this date time is to get the data as of a certain date - in our case we want the latest data - can remove this comment if that makes sense to you

app/services/CategorisationService.scala
```diff
@@ -39,122 +39,87 @@ class CategorisationService @Inject() (
...
-        )
-      updatedAnswersCleanUpAssessments <-
-        Future.fromTry(cleanupOldAssessmentAnswers(updatedAnswers, recordId))
-      _                                <- sessionRepository.set(updatedAnswersCleanUpAssessments)
-    } yield updatedAnswersCleanUpAssessments
-  }
+  ): Future[UserAnswers] =
+    (for {
+      getGoodsRecordResponse <- goodsRecordsConnector.getRecord(eori = request.eori, recordId = recordId)
+      newCommodityCode       <- Future.fromTry(Try(request.userAnswers.get(LongerCommodityQuery(recordId)).get))
```
liandra-giacomine, line 117: Can we handle the exception here/not do .get

app/controllers/CategoryGuidanceController.scala
ChrisCookOC: It only gets set if we go to the longer commodity code page, so if you are just normally categorising it won't be set.

app/models/CategoryRecord.scala
ChrisCookOC: We actually want the number of assessments they answered. Replaced it with that

app/models/ott/CategorisationInfo.scala
ChrisCookOC: Nice idea :) Done

app/models/ott/CategoryAssessment.scala
ChrisCookOC: Done

app/navigation/Navigator.scala
ChrisCookOC: I think it's fine. By which I mean it was already there and is there on the normal mode version too 😄 

Guess it's to avoid having to map the value. Probably should be going to JourneyRecovery if its not defined, but I'm kinda happy to say its existing technical debt and run away?

app/services/CategorisationService.scala
ChrisCookOC: Refactored

app/services/CategorisationService.scala
ChrisCookOC: Removed 🙂

app/services/CategorisationService.scala
ChrisCookOC: The Future.fromTry handles it doesn't it? - it will return a Future.failed if it isn't defined

Ewan-James-Donovan: I agree, I believe .get should be allowed in cases where it's handled safely

Ewan-James-Donovan:
Would approve after conflict fixes

Ewan-James-Donovan (approved):

PR #227, opened 2024-07-12
Route create record request to data store

liandra-giacomine:


app/models/router/responses/GetGoodsRecordResponse.scala
```diff
@@ -16,58 +16,94 @@
...
+  supplementaryUnit: Option[Double] = None,
+  measurementUnit: Option[String],
+  comcodeEffectiveFromDate: Instant,
+  comcodeEffectiveToDate: Option[Instant] = None,
+  version: Int,
+  active: Boolean,
+  toReview: Boolean,
+  reviewReason: Option[String] = None,
   declarable: String,
+  ukimsNumber: Option[String] = None,
```
AmarkumarJagtap, line 43: think ukimsNumber is still mandatory at the moment

ChrisCookOC (approved):

app/models/router/responses/GetGoodsRecordResponse.scala
liandra-giacomine: I copied from the router model which has it as optional

https://github.com/hmrc/trader-goods-profiles-router/blob/main/app/uk/gov/hmrc/tradergoodsprofilesrouter/models/response/CreateOrUpdateRecordResponse.scala#L46

PR #231, opened 2024-07-12
TGP-1265 added auditing for update goods record

AmarkumarJagtap:


app/controllers/HasCommodityCodeChangeController.scala
```diff
@@ -62,11 +66,19 @@ class HasCommodityCodeChangeController @Inject() (
         .bindFromRequest()
         .fold(
           formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, recordId))),
-          value =>
+          value => {
+            auditService
```
ChrisCookOC: Won't this trigger the audit if they select No, they don't want to set a commodity code? (Same question for all Has Controllers)

Ewan-James-Donovan (dismissed):

app/controllers/HasCommodityCodeChangeController.scala
AmarkumarJagtap: think this is start of audit event if they wish to modify when they click on change link and finish audit event with actual changed value from CyaUpdateGoodsRecordController when they submit the changes?

liandra-giacomine (approved):

ChrisCookOC (approved):

PR #237, opened 2024-07-15
TGP 1395: Success banner on goods record

vinoocs:


app/controllers/AdviceStartController.scala
```diff
@@ -40,7 +40,7 @@ class AdviceStartController @Inject() (
 
   def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData) {
     implicit request =>
-      Ok(view(recordId))
+      Ok(view(recordId)).removingFromSession("changesMade", "changedPage")
```
liandra-giacomine: Can we store these strings in a constants file maybe named SessionData?

app/controllers/CommodityCodeController.scala
```diff
@@ -107,14 +107,20 @@ class CommodityCodeController @Inject() (
...
-            (for {
-              commodity               <-
-                ottConnector.getCommodityCode(value, request.eori, request.affinityGroup, CreateRecordJourney, None)
-              updatedAnswers          <- Future.fromTry(request.userAnswers.set(CommodityCodeUpdatePage(recordId), value))
-              updatedAnswersWithQuery <-
-                Future.fromTry(updatedAnswers.set(CommodityUpdateQuery(recordId), commodity))
-              _                       <- sessionRepository.set(updatedAnswersWithQuery)
-            } yield Redirect(navigator.nextPage(CommodityCodeUpdatePage(recordId), mode, updatedAnswersWithQuery)))
+            {
+              val oldValue = request.userAnswers.get(CommodityCodeUpdatePage(recordId)).getOrElse("")
```
liandra-giacomine: Is it isn't set it will return an empty option, we wouldn't want to default it to an empty string here but instead map on the option and check if it matches

app/controllers/GoodsDescriptionController.scala
```diff
@@ -92,11 +93,15 @@ class GoodsDescriptionController @Inject() (
...
           formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, submitAction))),
-          value =>
+          value => {
+            val oldValue = request.userAnswers.get(GoodsDescriptionUpdatePage(recordId)).getOrElse("")
             for {
               updatedAnswers <- Future.fromTry(request.userAnswers.set(GoodsDescriptionUpdatePage(recordId), value))
               _              <- sessionRepository.set(updatedAnswers)
             } yield Redirect(navigator.nextPage(GoodsDescriptionUpdatePage(recordId), mode, updatedAnswers))
+              .addingToSession("changesMade" -> (oldValue != value).toString)
+              .addingToSession("changedPage" -> "goods description")
```
liandra-giacomine: Can we add these pages to the session data constants file too please? :)

app/controllers/AdviceStartController.scala
vinoocs: done it!

app/controllers/CommodityCodeController.scala
vinoocs: not defaulting to string and using option now

app/controllers/GoodsDescriptionController.scala
vinoocs: done!

liandra-giacomine (approved):

PR #244, opened 2024-07-16
TGP-1824 Send country of origin in category assessments call to Green lanes API

pauloncast:


ChrisCookOC:
Do we want to update the stub data to only be for one country as well? Currently it returns data for every country (might be easier not to if it doesn't affect anything on the frontend?)

Just had some questions about why sometimes the same category assessment is asked twice and its because its in the stub data twice for different countries!

Ewan-James-Donovan:
Would approve after conflicts are resolved

Ewan-James-Donovan (approved):

PR #250, opened 2024-07-17
TGP-1752: Display measurement unit

vinoocs:
- Displaying measurement unit received from OTT


yohan-oc (dismissed):

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
```diff
@@ -17,24 +17,32 @@
...
-          ActionItemViewModel("site.change", routes.SupplementaryUnitController.onPageLoad(CheckMode, recordId).url)
-            .withVisuallyHiddenText(messages("supplementaryUnit.change.hidden"))
-        )
-      )
+  def row(answers: UserAnswers, recordId: String)(implicit messages: Messages): Option[SummaryListRow] = {
+    val recordCategorisations = answers.get(RecordCategorisationsQuery).getOrElse(RecordCategorisations(Map.empty))
+    val categorisationInfoOpt = recordCategorisations.records.get(recordId)
+    categorisationInfoOpt match {
+      case Some(categorisationInfo) =>
+        val measurementUnit = categorisationInfo.measurementUnit.getOrElse("")
```
liandra-giacomine: do we really want to default it in this case? if it's not present we probably don't want to display anything for this

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
```diff
@@ -17,24 +17,32 @@
...
-        actions = Seq(
-          ActionItemViewModel("site.change", routes.SupplementaryUnitController.onPageLoad(CheckMode, recordId).url)
-            .withVisuallyHiddenText(messages("supplementaryUnit.change.hidden"))
-        )
-      )
+  def row(answers: UserAnswers, recordId: String)(implicit messages: Messages): Option[SummaryListRow] = {
+    val recordCategorisations = answers.get(RecordCategorisationsQuery).getOrElse(RecordCategorisations(Map.empty))
+    val categorisationInfoOpt = recordCategorisations.records.get(recordId)
+    categorisationInfoOpt match {
+      case Some(categorisationInfo) =>
```
liandra-giacomine: what do we do if the option is empty? Journey recovery?

vinoocs: yeah I had that question in mind too, is it okay to redirect to journey recovery?

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
Ewan-James-Donovan: `answer + " " + measurementUnit` I think it'd look fine visually since it'll just append nothing here.  If you wanted to be safe, the option should determine whether that string has a space and includes the measurementUnit

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
Ewan-James-Donovan: I believe it should do that automatically but yes JourneyRecovery is where it should go.

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
vinoocs: done it!

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
vinoocs: added a test case to make sure if it redirects journeyrecovery

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
```diff
@@ -17,24 +17,35 @@
...
+        val measurementUnit = categorisationInfo.measurementUnit
+        answers.get(SupplementaryUnitPage(recordId)).map { answer =>
+          val value = if (measurementUnit.nonEmpty) s"$answer ${measurementUnit.get.trim}" else answer
```
liandra-giacomine: Do we want a summary row if the value is empty? I think this should work (SUDO CODE):


```

(for {
recordCategorisations  <- answers.get(RecordCategorisationsQuery)
categorisationInfo <- recordCategorisations.records.get(recordId)
measurementUnit <- categorisationInfo.measurementUnit
supplementaryUnit <- answers.get(SupplementaryUnitPage(recordId))
} yield supplementaryUnit).map{value => 
SummaryListRowViewModel(
             key = "supplementaryUnit.checkYourAnswersLabel",
             value = ValueViewModel(value),
             actions = Seq(
               ActionItemViewModel("site.change", routes.SupplementaryUnitController.onPageLoad(CheckMode, recordId).url)
                 .withVisuallyHiddenText(messages("supplementaryUnit.change.hidden"))
             )
           )}
```

vinoocs: In my mind, we have that logic handled before this piece of code is being called - we get to this page only when we have a value for supplementary unit - otherwise we won't show this row at all. do we need to handle it again here

vinoocs: have updated to use for comprehension now

liandra-giacomine (approved):

PR #262, opened 2024-07-19
TGP-1812: Improve logging

yohan-oc:


app/controllers/RemoveNirmsController.scala
```diff
@@ -88,7 +88,7 @@ class RemoveNirmsController @Inject() (
     val errorMessages = errors.toChain.toList.map(_.message).mkString(", ")
 
     val continueUrl = RedirectUrl(routes.HasNirmsController.onPageLoadUpdate.url)
-    logger.warn(s"Unable to update Trader profile.  Missing pages: $errorMessages")
+    logger.error(s"Unable to update Trader profile.  Missing pages: $errorMessages")
```
AmarkumarJagtap, line 91: Think we need to check if they raise any pager duty alerts now if we change it to error

yohan-oc: We are good here

AmarkumarJagtap (approved):

Ewan-James-Donovan (approved):

liandra-giacomine (approved):

PR #264, opened 2024-07-19
Various Longer Comcode Bugfixes

Ewan-James-Donovan:
Greatly simplified the code.

https://jira.tools.tax.service.gov.uk/browse/TGP-1896
- Answer is preserved after back button is pressed.

https://jira.tools.tax.service.gov.uk/browse/TGP-1898
- Now only redirects in CheckMode if the comcode is the same as was provided last time (As per TGP-1430).
- Otherwise the journey will continue to has correct goods, then if recategorisation is needed it will show those questions, if not it will go to CYA (Please note this is the desired behaviour, or at least behaviour I had no part coding. Please message before giving back to dev as Chris and I may want to speak about it).

https://jira.tools.tax.service.gov.uk/browse/TGP-1899
- Answer is preserved as user provided it with no modifications

https://jira.tools.tax.service.gov.uk/browse/TGP-1908
- Boolean value for HasCorrectGoods is cleared when the user submits a new comcode

https://jira.tools.tax.service.gov.uk/browse/TGP-1871
- Longer comcode is only shown when the user has provided it.


conf/application.conf
```diff
@@ -120,7 +120,7 @@ enrolment-config {
 }
 
 auditing {
-  enabled = true
+  enabled = false
```
ChrisCookOC: Shouldn't commit this!

Ewan-James-Donovan: Oops, you're absolutely right. Fixed

ChrisCookOC (dismissed):

vinoocs (dismissed):

app/controllers/LongerCommodityCodeController.scala
```diff
@@ -52,79 +51,74 @@ class LongerCommodityCodeController @Inject() (
...
-          form
-      }
-
-      originalComcodeOpt match {
-        case Some(shortCommodity) if commodityCodeSansTrailingZeros(shortCommodity).length == 6 =>
-          Ok(view(preparedForm, mode, commodityCodeSansTrailingZeros(shortCommodity), recordId))
-        case _                                                                                  =>
+      val previousAnswerOpt = request.userAnswers.get(LongerCommodityCodePage(recordId))
+      val shortComcodeOpt   = getShortCommodityCodeOpt(recordId, request, validLength)
+      val preparedForm      = previousAnswerOpt.map(answer => form.fill(answer)).getOrElse(form)
```
besscerule: Can we use this pattern so it matches other pages

`      val preparedForm = request.userAnswers.get(LongerCommodityCodePage(recordId)) match {
        case None        => form
      }`

app/controllers/LongerCommodityCodeController.scala
```diff
@@ -52,79 +51,74 @@ class LongerCommodityCodeController @Inject() (
...
             .fold(
-              formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, shortCode, recordId))),
+              formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, shortComcode, recordId))),
               value => {
-                val longCommodityCode   = s"$shortCode$value"
-                val shouldRedirectToCya = latestComcodeOpt.exists(latest =>
-                  shortCode.concat(value) == latest &&
-                    categorisationInfoOpt.exists(_.latestDoesNotMatchOriginal)
+                val longCommodityCode   = (shortComcode + value).padTo(maxValidLength, "0").mkString
+                val shouldRedirectToCya = currentlyCategorisedOpt.contains(longCommodityCode) && mode == CheckMode
```
besscerule, line 93: this feels like it should just be comparing the value with what's already in userAnswers why are you looking at the recordCategorisations query?

app/controllers/LongerCommodityCodeController.scala
```diff
@@ -137,26 +131,26 @@ class LongerCommodityCodeController @Inject() (
                                    Some(recordId)
                                  )
       updatedAnswers          <- Future.fromTry(request.userAnswers.set(LongerCommodityCodePage(recordId), value))
-      updatedAnswersWithQuery <- Future.fromTry(updatedAnswers.set(LongerCommodityQuery(recordId), validCommodityCode))
+      cleansedAnswers         <- if (shouldRedirect) {
+                                   Future.successful(updatedAnswers)
+                                 } else {
+                                   Future.fromTry(updatedAnswers.remove(HasCorrectGoodsLongerCommodityCodePage(recordId)))
```
besscerule: i think this should be in the LongerCommodityCodePage file - if you check the one for the plain commodityCodePage you can see it has a cleanup for this

besscerule (commented):
Some bits to look at

app/controllers/LongerCommodityCodeController.scala
Ewan-James-Donovan: Because it's possible that a user is in CheckMode, changes their answer, then selects "no" on HasCorrectGoods, comes back and then just submits that answer again, redirecting them to CYA with a different answer that they didn't ever say yes to. The source of truth for commodity codes that have been or are ready to be categorised is recordCategorisations, which is updated with the answer the user puts in if they say "yes". I hope that makes sense?

app/controllers/LongerCommodityCodeController.scala
Ewan-James-Donovan: Sure :)

app/controllers/LongerCommodityCodeController.scala
Ewan-James-Donovan: Agreed

Ewan-James-Donovan: Actually, cleanup does not support recordId and that is needed for the cleanup method to reference the pervious value... had to undo the change...

ChrisCookOC: Cleanup doesn't need to pass in recordId, it is in scope as its in the Page definition. 

See e.g. HasSupplementaryUnitPage cleanup

Ewan-James-Donovan: oof, yep you're right, I was so close on my initial try :') Done

yohan-oc (dismissed):

vinoocs:
Looks good to me, just that we shouldn't be using CountryOfOriginPageJourney anymore as it wouldn't work if we go through view records journey. But I see Chris has raised a PR for fixing that so it should be fine :).. Okay Chris's PR is merged so it needs to be pulled now in this PR.

vinoocs (dismissed):

liandra-giacomine (dismissed):

yohan-oc (approved):

app/controllers/LongerCommodityCodeController.scala
```diff
@@ -53,79 +52,77 @@ class LongerCommodityCodeController @Inject() (
...
-  private def saveAnswerAndProceedWithJourney(
+  private def getShortCommodityCodeOpt(
+    recordId: String,
+    request: DataRequest[AnyContent],
+    validLength: Int
+  ): Option[String] =
+    request.userAnswers
+      .get(RecordCategorisationsQuery)
+      .flatMap(_.records.get(recordId))
+      .flatMap(_.originalCommodityCode.map(_.reverse.dropWhile(_ == "0").reverse.padTo(validLength, "0").mkString))
```
liandra-giacomine, line 117: what about 6 digit commodity codes that end in a 0, wouldn't this make it 5 for example?

Ewan-James-Donovan: thankfully the answer is no

`.padTo(validLength, "0")`

ensures it is 6

app/controllers/LongerCommodityCodeController.scala
```diff
@@ -53,79 +52,77 @@ class LongerCommodityCodeController @Inject() (
...
-      val categorisationInfoOpt = request.userAnswers.get(RecordCategorisationsQuery).flatMap(_.records.get(recordId))
-
-      val originalComcodeOpt = categorisationInfoOpt.flatMap(_.originalCommodityCode)
-      val latestComcodeOpt   = categorisationInfoOpt.map(_.commodityCode)
-
-      originalComcodeOpt match {
-        case Some(shortCommodity) if commodityCodeSansTrailingZeros(shortCommodity).length == 6 =>
-          val shortCode = commodityCodeSansTrailingZeros(shortCommodity)
+      val shortComcodeOpt         = getShortCommodityCodeOpt(recordId, request, validLength)
+      val currentlyCategorisedOpt =
```
besscerule, line 79: You have a few options floating about that you use lower down in the code. To make it more functional it might be worth refactoring so you don't have these gaps and instead transform the data to what you wish to return

Tbf it looks like it works so I don't care if it stays like this lol

PR #265, opened 2024-07-19
TGP- 1856/recategorisation - don't redisplay old questions

mycp98:


Ewan-James-Donovan (dismissed):

Ewan-James-Donovan (dismissed):

app/controllers/AssessmentController.scala
```diff
@@ -64,16 +65,30 @@ class AssessmentController @Inject() (
...
+          case Some(_)              => true
+          case None                 => false
+        }
         if (exemptions.isEmpty) {
           Future.successful(
             Redirect(
               navigator.nextPage(AssessmentPage(recordId, index, shouldRedirectToCya = true), mode, request.userAnswers)
             )
           )
+        } else if (areWeRecategorising && hasAssessmentBeenAnswered && mode == NormalMode) {
```
ChrisCookOC, line 80: We'll probably be able to refactor this as part of the upcoming refactoring. Something to bear in mind with that :)

app/controllers/AssessmentController.scala
```diff
@@ -64,16 +65,30 @@ class AssessmentController @Inject() (
                                             case Some(value) => form.fill(value)
                                             case None        => form
                                           }
-      } yield
+      } yield {
+        val areWeRecategorising = request.userAnswers.get(RecategorisingQuery(recordId)).getOrElse(false)
+
```
liandra-giacomine, line 70: I think this can be updated to

`val hasAssessmentBeenAnswered = request.userAnswers.get(AssessmentPage(recordId, index)).exists(_ != NotAnsweredYet)`

app/controllers/CategorisationResultController.scala
```diff
@@ -18,24 +18,37 @@ package controllers
...
-) extends FrontendBaseController
+  view: CategorisationResultView,
+  sessionRepository: SessionRepository
+)(implicit ec: ExecutionContext)
+    extends FrontendBaseController
     with I18nSupport {
 
   def onPageLoad(recordId: String, scenario: Scenario): Action[AnyContent] =
     (identify andThen getData andThen requireData) { implicit request =>
+      if (request.userAnswers.get(RecategorisingQuery(recordId)).isDefined) {
```
liandra-giacomine: what do we do if it's not defined?

app/controllers/AssessmentController.scala
ChrisCookOC: oooh that's clever. Let's give it a try 🙂

ChrisCookOC: Oh was hoping it would kill the recategorising query too, but don't think it can 😢

app/controllers/CategorisationResultController.scala
mycp98: i've updated this now, removed the if statement as the functionality is the same without it

yohan-oc (approved):

vinoocs (approved):

PR #272, opened 2024-07-22
TGP-1391 - ensure country of origin comes from goods record

ChrisCookOC:


test/controllers/LongerCommodityCodeControllerSpec.scala
```diff
@@ -69,271 +69,292 @@ class LongerCommodityCodeControllerSpec extends SpecBase with MockitoSugar {
...
+
+            val result = route(application, request).value
+
+            status(result) mustEqual BAD_REQUEST
+            contentAsString(result) mustEqual view(boundForm, NormalMode, shortCommodity, testRecordId)(
+              request,
+              messages(application)
+            ).toString
+
+            verify(mockOttConnector, times(1)).getCommodityCode(any(), any(), any(), any(), any(), any())(any())
```
AmarkumarJagtap: may be I think we should start removing times(1) from newly added tests as its already defaulted times(1) so we don't need to explicitly add it here. :)

ChrisCookOC: Fair 🙂

Ewan-James-Donovan (approved):
LGTM :)

vinoocs (approved):

PR #275, opened 2024-07-23
TGP-1840: QA does not display correct layout - Two-thirds column

mycp98:


vinoocs (approved):

AmarkumarJagtap (approved):

liandra-giacomine (approved):

app/views/components/fullWidthMainContent.scala.html
```diff
@@ -19,7 +19,7 @@
 @(contentBlock: Html)
 
 <div class="govuk-grid-row">
- <div class="govuk-grid-column-full">
+ <div class="govuk-grid-column-two-thirds">
   @contentBlock
```
yohan-oc, line 23: image
This makes the table layout two third. Issue is with the length of the word.

yohan-oc (changes requested):

app/views/components/fullWidthMainContent.scala.html
mycp98: Thanks, yeah after speaking with Lydia the table needs to be full-width, so I think the way it was before was correct

PR #278, opened 2024-07-23
TGP1808 : add supplementary unit on goods record

vinoocs:


yohan-oc (approved):

app/viewmodels/checkAnswers/HasSupplementaryUnitSummary.scala
```diff
@@ -39,4 +39,20 @@ object HasSupplementaryUnitSummary {
...
+
+  def row(value: Boolean, recordId: String)(implicit messages: Messages): Option[SummaryListRow] = {
+
+    val textValue = if (value) "site.yes" else "site.no"
+    Some(
+      SummaryListRowViewModel(
+        key = "hasSupplementaryUnit.checkYourAnswersLabel",
+        value = ValueViewModel(textValue),
+        actions = Seq(
+          ActionItemViewModel("site.change", routes.HasSupplementaryUnitController.onPageLoad(CheckMode, recordId).url)
```
ChrisCookOC, line 51: Is going through check mode right? It doesn't work currently either way 🙈 if you select Yes it will load the SupplementaryUnit entry page, and if you select No it will load the CyaCategorisation, but both are failing to load in the case where I just log in and go straight to the view records and try to update.

ChrisCookOC (changes requested):

app/viewmodels/checkAnswers/HasSupplementaryUnitSummary.scala
vinoocs: You're right, that's exactly was I was talking to Amaka today. There's a separate ticket to implement the functionality of the change links. So we haven't done in this one.

Ewan-James-Donovan (approved):

app/viewmodels/checkAnswers/HasSupplementaryUnitSummary.scala
Ewan-James-Donovan: Approving as it looks like there's a way forward here

ChrisCookOC (approved):

PR #286, opened 2024-07-25
TGP-2013: PRA User profile check

yohan-oc:
Ticket https://jira.tools.tax.service.gov.uk/browse/TGP-2013

This may seem like a lot of changes, but the main idea is to prevent URL hacking without a user profile.

For example: Prevent users from accessing the home page without a profile, creating records, and accessing the goods record page.

I have created `ProfileAuthenticateAction` and updated the controller actions and unit tests.

That's it! :)

(I also ran the journey tests locally. All journey tests passed.)

### DIFF
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -26,7 +26,6 @@ import pages.CategoryGuidancePage
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import queries.RecordCategorisationsQuery
-import repositories.SessionRepository
 import services.{AuditService, CategorisationService}
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import utils.SessionData.{dataUpdated, pageUpdated}
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -40,20 +40,20 @@ class CategoryGuidanceController @Inject() (
   identify: IdentifierAction,
   getData: DataRetrievalAction,
   requireData: DataRequiredAction,
+  profileAuth: ProfileAuthenticateAction,
   val controllerComponents: MessagesControllerComponents,
   view: CategoryGuidanceView,
   auditService: AuditService,
   categorisationService: CategorisationService,
   navigator: Navigator,
-  goodsRecordConnector: GoodsRecordConnector,
-  sessionRepository: SessionRepository
+  goodsRecordConnector: GoodsRecordConnector
 )(implicit ec: ExecutionContext)
     extends FrontendBaseController
     with I18nSupport
     with Logging {
 
-  def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData).async {
-    implicit request =>
+  def onPageLoad(recordId: String): Action[AnyContent] =
+    (identify andThen profileAuth andThen getData andThen requireData).async { implicit request =>
       categorisationService
         .requireCategorisation(request, recordId)
         .flatMap { userAnswers =>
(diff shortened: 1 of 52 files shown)


test/controllers/RemoveNiphlControllerSpec.scala
```diff
@@ -82,8 +84,6 @@ class RemoveNiphlControllerSpec extends SpecBase with MockitoSugar {
 
       when(mockSessionRepository.set(any())) thenReturn Future.successful(true)
 
-      val mockTraderProfileConnector = mock[TraderProfileConnector]
```
ChrisCookOC, line 85: is mockTraderProfileConnector being reset anywhere? 
Assuming this had its own one to avoid any other test affecting the verify on line 106?

app/controllers/CategoryGuidanceController.scala
```diff
@@ -40,20 +40,20 @@ class CategoryGuidanceController @Inject() (
...
   requireData: DataRequiredAction,
+  profileAuth: ProfileAuthenticateAction,
   val controllerComponents: MessagesControllerComponents,
   view: CategoryGuidanceView,
   auditService: AuditService,
   categorisationService: CategorisationService,
   navigator: Navigator,
-  goodsRecordConnector: GoodsRecordConnector,
-  sessionRepository: SessionRepository
+  goodsRecordConnector: GoodsRecordConnector
```
vinoocs, line 48: can we also remove the import for SessionRepository

app/controllers/NameController.scala
```diff
@@ -46,18 +47,18 @@ class NameController @Inject() (
...
         case Some(value) => form.fill(value)
       }
 
       Ok(view(preparedForm, mode, recordId))
-  }
+    }
 
   def onSubmit(mode: Mode, recordId: String): Action[AnyContent] =
-    (identify andThen getData andThen requireData).async { implicit request =>
+    (identify andThen profileAuth andThen getData andThen requireData).async { implicit request =>
```
vinoocs, line 61: is it necessary to have this on submit as we check on pageload anyway?

app/controllers/PreviousMovementRecordsController.scala
```diff
@@ -32,27 +32,30 @@ class PreviousMovementRecordsController @Inject() (
...
+            case true  => Redirect(routes.GoodsRecordsController.onPageLoad(firstPage))
+          }
+      }
   }
 
-  def onSubmit: Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
-    goodsRecordConnector
-      .storeAllRecords(request.eori)
-      .map(_ => Redirect(routes.GoodsRecordsController.onPageLoad(firstPage)))
+  def onSubmit: Action[AnyContent] = (identify andThen profileAuth andThen getData andThen requireData).async {
```
vinoocs, line 55: same as above

test/controllers/RemoveNiphlControllerSpec.scala
yohan-oc: No. I removed this because it was a duplicate, already declared on line 50. Do you think this needs to be revert?

app/controllers/CategoryGuidanceController.scala
yohan-oc: Removed

test/controllers/RemoveNiphlControllerSpec.scala
ChrisCookOC: If it works I'm happy. Was just trying to figure out the reason it was declared separately beforehand.

app/controllers/NameController.scala
yohan-oc: This is to prevent making POST requests using third-party tools.

app/controllers/PreviousMovementRecordsController.scala
yohan-oc: Here also the same.  Prevent making POST requests using third-party tools

ChrisCookOC (approved):

vinoocs (approved):

PR #291, opened 2024-07-30
TGP-2016 Redirect to Longer Commodity Code when Cat2 no exemp and longer comcode is needed

Ewan-James-Donovan:
Here's what a user might do:

- Create record
- Categorise now
- Answer all questions prior to the category 2 assessment without an exemption
- IF AND ONLY IF there are descendants and the commodity code is short (in other words, if the current code is NOT the bottom level) then redirect them to enter a longer commodity code.

Please note:

- There is stub data alongside this work: https://github.com/hmrc/trader-goods-profiles-stubs/pull/95 (Comcodes 170491 & 17049199)
- Since we will refactor next sprint, I have abandoned attempts to butcher the Navigator once again. I think this is fine since it will all be replaced.

### DIFF
--- a/app/controllers/AssessmentController.scala
+++ b/app/controllers/AssessmentController.scala
@@ -76,15 +76,24 @@ class AssessmentController @Inject() (
 
         val commodityCodeWithoutZeros = commodityCode.reverse.dropWhile(char => char == '0').reverse
 
-        val shouldGoToLongerCommodityCode = exemptions.isEmpty && category == 2 && commodityCodeWithoutZeros.length <= 6 &&
-          categorisationInfo.descendantCount != 0
+        val shouldGoToLongerCommodityCode =
+          exemptions.isEmpty && category == 2 && commodityCodeWithoutZeros.length <= 6 &&
+            categorisationInfo.descendantCount != 0
 
         val shouldGoToCya = exemptions.isEmpty && !shouldGoToLongerCommodityCode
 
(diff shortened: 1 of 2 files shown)


app/controllers/AssessmentController.scala
```diff
@@ -71,20 +72,29 @@ class AssessmentController @Inject() (
...
+
+        val commodityCodeWithoutZeros = commodityCode.reverse.dropWhile(char => char == '0').reverse
+
+        val shouldGoToLongerCommodityCode =
+          exemptions.isEmpty && category == 2 && commodityCodeWithoutZeros.length <= 6 &&
+            categorisationInfo.descendantCount != 0
+
+        val shouldGoToCya = exemptions.isEmpty && !shouldGoToLongerCommodityCode
+
+        (shouldGoToCya, shouldGoToLongerCommodityCode, shouldDisplayNext) match {
```
Ewan-James-Donovan, line 85: If this looks less than optimal, please consider it is very likely to be refactored or replaced in the coming sprint.

ChrisCookOC: Yes it definitely is not going to make the cut 😆

Ewan-James-Donovan: Exactly :')

ChrisCookOC (approved):
there's tests which is all that really matters to me right now. Like you say, the code itself will be improved upon goign forward

PR #297, opened 2024-07-31
TGP-1694: value for Category row fix

mycp98:


app/controllers/SingleRecordController.scala
```diff
@@ -74,9 +74,14 @@ class SingleRecordController @Inject() (
           )
         )
 
+        val categoryValue         = record.category.toString match {
+          case "1" => "Category 1"
```
AmarkumarJagtap: does it needs to be string match? looks like they are integers

AmarkumarJagtap (dismissed):

app/controllers/SingleRecordController.scala
mycp98: thanks 😄 updated now

ChrisCookOC (approved):

AmarkumarJagtap (approved):

PR #302, opened 2024-08-01
TGP-1961 + Others. All review reason pages.

Ewan-James-Donovan:


app/controllers/ReviewReasonController.scala
```diff
@@ -0,0 +1,56 @@
...
+)(implicit ec: ExecutionContext)
+    extends FrontendBaseController
+    with I18nSupport {
+
+  def onPageLoad(recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData).async {
+    implicit request =>
+      goodsRecordConnector
+        .getRecord(request.eori, recordId)
+        .map { record =>
+          record.reviewReason match {
```
vinoocs, line 47: are we translating reviewReason somewhere. This is what I see in the stub  "reviewReason":"The commodity code has expired. You'll need to change the commodity code and categorise the goods.",

Ewan-James-Donovan: Just spoken with Liandra and Ben, there's some debate on whether what the frontend gets in the goodsRecord is the longer review reason or shorter reason. For now, review as normal.

NO MERGING tho :)

app/views/GoodsRecordsSearchResultView.scala.html
```diff
@@ -84,7 +84,7 @@ <h1 class="govuk-heading-l">@messages("goodsRecordSearchResult.h1", totalRecords
                                 text2 = messages("site.remove"),
                                 visuallyHiddenContent1 = messages("goodsRecords.hidden.change"),
                                 visuallyHiddenContent2 = messages("goodsRecords.hidden.remove"),
-                                href1 = routes.SingleRecordController.onPageLoad(goodsRecord.recordId).url,
+                                href1 = routes.ReviewReasonController.onPageLoad(goodsRecord.recordId).url,
```
besscerule, line 87: if this needs to be wherever the single page pops up might it be better if this was on the single goods record onPageLoad, which redirects to this page if necessary

this just means other links will also show it - like when you have just created a good but don't want to categorise it yet

app/controllers/ReviewReasonController.scala
Ewan-James-Donovan: The shorter reason will be used on the frontend side, what's displayed to the user is defined in the messages file, however. It's sort of translated within the messages file.

app/views/GoodsRecordsSearchResultView.scala.html
Ewan-James-Donovan: I agree - done :)

vinoocs (dismissed):

app/controllers/SingleRecordController.scala
```diff
@@ -67,57 +67,60 @@ class SingleRecordController @Inject() (
...
-        val detailsList = SummaryListViewModel(
-          rows = Seq(
-            TraderReferenceSummary.row(record.traderRef, recordId, NormalMode),
-            GoodsDescriptionSummary.row(record.goodsDescription, recordId, NormalMode),
-            CountryOfOriginSummary.row(record.countryOfOrigin, recordId, NormalMode),
-            CommodityCodeSummary.row(record.comcode, recordId, NormalMode),
-            StatusSummary.row(record.declarable)
+      } yield
+        if (record.toReview) {
+          Redirect(routes.ReviewReasonController.onPageLoad(recordId).url)
```
ChrisCookOC: Bit wary of us doing this in here, because its the kind of thing we want to avoid doing in our controllers. Ideally the navigator needs to figure out it needs to show that page instead of this one? 

(which admittedly might be complicated as this page is linked from places directly I guess)

app/views/GoodsRecordsSearchResultView.scala.html
Ewan-James-Donovan: Unfortunately, I've had to undo this change. The issue is the only way to tell if we need to show the review page is through the toReview field... If we send them to Single record and decide whether to show it there, it would end up in an infinite loop on showing the review page.

Ewan-James-Donovan: On second thoughts, It will also only be needed here.

app/controllers/SingleRecordController.scala
Ewan-James-Donovan: Have a look at the review reason controller, this has been undone - It's a single redirect that happens on page load as well... I think pulling that out to the navigator might be overkill + it's based on a link and nothing else.

vinoocs (approved):

yohan-oc (approved):

PR #309, opened 2024-08-06
TGP-1870: Fixed values getting reset in backlink

vinoocs:
- The original values need to be set to session were reset on backlink - it is now fixed
- Improved test coverage
- converting supplementary unit to bigdecimal for accurate comparison of values

### DIFF
--- a/app/controllers/CyaSupplementaryUnitController.scala
+++ b/app/controllers/CyaSupplementaryUnitController.scala
@@ -84,20 +84,19 @@ class CyaSupplementaryUnitController @Inject() (
       SupplementaryRequest.build(request.userAnswers, request.eori, recordId) match {
         case Right(model) =>
           //TODO : Audit service implementation
-          val (initialHasSuppUnitOpt, initialSuppUnitOpt) = (
-            request.session.get(initialValueOfHasSuppUnit).map(_.toBoolean),
-            request.session.get(initialValueOfSuppUnit)
-          )
-          val (finalHasSuppUnit, finalSuppUnit)           = (
-            model.hasSupplementaryUnit.getOrElse(false),
-            model.supplementaryUnit.getOrElse("")
-          )
-          val isValueChanged                              =
+
+          val initialHasSuppUnitOpt = request.session.get(initialValueOfHasSuppUnit).map(_.toBoolean)
+          val initialSuppUnitOpt    = request.session.get(initialValueOfSuppUnit)
+
+          val finalHasSuppUnit = model.hasSupplementaryUnit.getOrElse(false)
+          val finalSuppUnit    = model.supplementaryUnit.getOrElse("")
+
+          val isValueChanged    =
             initialHasSuppUnitOpt.exists(_ != finalHasSuppUnit) || compareSupplementaryUnits(
               initialSuppUnitOpt.getOrElse("0"),
               finalSuppUnit
             )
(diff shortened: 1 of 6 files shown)


app/controllers/CyaSupplementaryUnitController.scala
```diff
@@ -93,7 +93,10 @@ class CyaSupplementaryUnitController @Inject() (
             model.supplementaryUnit.getOrElse("")
```
ChrisCookOC: (slightly above) why are these things being defined together rather than separately. 
val initialHasSuppUnitOpt = request.session.get(initialValueOfHasSuppUnit).map(_.toBoolean)
val initialSuppUnitOpt = request.session.get(initialValueOfSuppUnit)

I guess it works but it makes the code more complicated, since they aren't being used as tuples?

vinoocs: I initially defined them separately, but it looked there were more repeated lines and val declarations just for getting the values, so simplified it :). If that makes it look complicated, then I will declare them separately.

ChrisCookOC: Ok, that's fair, although its more lines of code to do it as tuples than just variables 🙂

app/controllers/CyaSupplementaryUnitController.scala
```diff
@@ -93,7 +93,10 @@ class CyaSupplementaryUnitController @Inject() (
...
           )
           val isValueChanged                              =
-            initialHasSuppUnitOpt.exists(_ != finalHasSuppUnit) || initialSuppUnitOpt.exists(_ != finalSuppUnit)
+            initialHasSuppUnitOpt.exists(_ != finalHasSuppUnit) || compareSupplementaryUnits(
```
ChrisCookOC: I'm think I'm finding the mix of getOrElse on some options and just using some options directly a bit confusing to follow! 

Any reason to not leave them all as options? Something like
` val initialHasSuppUnitOpt2 =             request.session.get(initialValueOfHasSuppUnit).map(_.toBoolean)

          val initialSuppUnitOpt2 =             request.session.get(initialValueOfSuppUnit)

          val finalhasSuppUnitOpt2 = model.hasSupplementaryUnit
          val finalSuppUnitOpt2 = model.supplementaryUnit
          
                    val isValueChangedd = initialHasSuppUnitOpt2 == finalhasSuppUnitOpt2 || compareSupplementaryUnits2(initialSuppUnitOpt2, finalSuppUnitOpt2)`

with

`  private def compareSupplementaryUnits2(initialSuppUnit: Option[String], finalSuppUnit: Option[String]): Boolean = {
    val initialSuppUnitBD = BigDecimal(initialSuppUnit.getOrElse("0"))
    val finalSuppUnitBD = BigDecimal(finalSuppUnit.getOrElse("0"))
    initialSuppUnitBD != finalSuppUnitBD
  }`
  
  Dunno, let me know what you think.

vinoocs: Makes sense to be consistent, modified!

app/controllers/CyaSupplementaryUnitController.scala
vinoocs: Done it to be more readable.

ChrisCookOC (approved):

PR #313, opened 2024-08-07
TGP-2125: supplementaryunit datatype inconsistency

vinoocs:
- The datatype of supplementary unit in API is Bigdecimal & in UI it is Double so making it consistent. 
- Also, it needs to be BigDecimal for displaying it in proper format in the UI


yohan-oc (dismissed):

ChrisCookOC (dismissed):

app/controllers/CyaSupplementaryUnitController.scala
```diff
@@ -91,11 +92,16 @@ class CyaSupplementaryUnitController @Inject() (
...
             initialHasSuppUnitOpt != finalHasSuppUnitOpt ||
               compareSupplementaryUnits(initialSuppUnitOpt, finalSuppUnitOpt)
+
+          val finalSuppUnitBD = convertToBigDecimal(finalSuppUnitOpt)
+
+          //if supplementary unit is zero, we consider it removed. This needs to be re-factored when API starts supporting null for removing objects
           val isSuppUnitRemoved =
-            initialHasSuppUnitOpt.contains(true) && finalHasSuppUnitOpt.contains(false)
+            (initialHasSuppUnitOpt
+              .contains(true) && finalHasSuppUnitOpt.contains(false)) || finalSuppUnitBD.contains(0)
```
AmarkumarJagtap, line 104: think we can do finalSuppUnitBD == 0 instead of contains as its already BigDecimal?

ChrisCookOC (dismissed):

yohan-oc (dismissed):

app/controllers/CyaSupplementaryUnitController.scala
vinoocs: Makes sense but can't compare BigDecimal with Int here, we should either have to use compareTo or equals as it is an object so keeping it like this would be more straightforward I believe. happy to change if you feel we could use compareTo/equals

AmarkumarJagtap (dismissed):

AmarkumarJagtap (approved):

PR #329, opened 2024-08-09
TGP-2055 No CDS enrolment unauthorised page.

Ewan-James-Donovan:
The old unauthorised page behaves as it should, this one is to cover when the enrolment for HMRC-CUS-ORG is not present.

### DIFF
--- a/app/controllers/actions/IdentifierAction.scala
+++ b/app/controllers/actions/IdentifierAction.scala
@@ -65,19 +65,26 @@ class AuthenticatedIdentifierAction @Inject() (
               checkUserAllowList(enrolment.value)(hc).flatMap { _ =>
                 block(IdentifierRequest(request, internalId, enrolment.value, affinityGroup))
               }
+            case Some(enrolment) if enrolment.value.isBlank  =>
+              throw InternalError("EORI is empty")
             case _                                           =>
               throw InsufficientEnrolments("Unable to retrieve Enrolment")
           }
       } recover {
-      case _: UserNotAllowedException =>
+      case _: UserNotAllowedException        =>
         logger.info("trader is not on user-allow-list redirecting to UnauthorisedServiceController")
         Redirect(routes.UnauthorisedServiceUserController.onPageLoad())
-      case _: NoActiveSession         =>
+      case _: NoActiveSession                =>
         logger.info(s"No Active Session. Redirect to $config.loginContinueUrl")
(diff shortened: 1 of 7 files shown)


app/controllers/actions/IdentifierAction.scala
```diff
@@ -65,18 +65,25 @@ class AuthenticatedIdentifierAction @Inject() (
               checkUserAllowList(enrolment.value)(hc).flatMap { _ =>
                 block(IdentifierRequest(request, internalId, enrolment.value, affinityGroup))
               }
+            case Some(enrolment) if enrolment.value.isBlank  =>
+              throw InternalError("EORI is empty")
```
ChrisCookOC, line 69: Which recovery case does this end up in?

Ewan-James-Donovan: It falls into the AuthorisationException, perhaps I should try using that instead of InternalError

Ewan-James-Donovan: Upon looking @ChrisCookOC this is the correct one to use:

```
object AuthorisationException {

  def fromString(reason: String): AuthorisationException = reason match {
    case "InsufficientConfidenceLevel" => new InsufficientConfidenceLevel
    case "UnsupportedAffinityGroup"    => new UnsupportedAffinityGroup
    case "UnsupportedCredentialRole"   => new UnsupportedCredentialRole
    case "UnsupportedAuthProvider"     => new UnsupportedAuthProvider
    case "BearerTokenExpired"          => new BearerTokenExpired
    case "MissingBearerToken"          => new MissingBearerToken
    case "InvalidBearerToken"          => new InvalidBearerToken
    case "SessionRecordNotFound"       => new SessionRecordNotFound
    case "IncorrectCredentialStrength" => new IncorrectCredentialStrength
    case "InsufficientEnrolments"      => new InsufficientEnrolments
    case "FailedRelationship"          => new FailedRelationship
    case "IncorrectNino"               => IncorrectNino
    case other                         => InternalError(other) <<<<<<<<<<<<<<<<<
  }
}
```

AuthorisationException class is abstract and can't be instantiated, also.

Ewan-James-Donovan: Ready for review

ChrisCookOC (approved):

mycp98 (approved):

PR #332, opened 2024-08-09
TGP-1946: Handle "Category" on Goods record page

yohan-oc:


app/controllers/CyaUpdateRecordController.scala
```diff
@@ -55,21 +55,24 @@ class CyaUpdateRecordController @Inject() (
...
   def onPageLoadCountryOfOrigin(recordId: String): Action[AnyContent] =
     (identify andThen getData andThen requireData).async { implicit request =>
-      UpdateGoodsRecord.buildCountryOfOrigin(request.userAnswers, request.eori, recordId) match {
-        case Right(_)     =>
-          val onSubmitAction = routes.CyaUpdateRecordController.onSubmitCountryOfOrigin(recordId)
-          getCountryOfOriginAnswer(request.userAnswers, recordId).map { answer =>
-            val list = SummaryListViewModel(
-              Seq(
-                CountryOfOriginSummary.row(answer, recordId, CheckMode)
+      goodsRecordConnector.getRecord(request.eori, recordId).flatMap { recordResponse =>
```
ChrisCookOC: what happens if the getRecord call fails?

app/models/UpdateGoodsRecord.scala
```diff
@@ -39,18 +39,18 @@ object UpdateGoodsRecord {
...
       Right(recordId),
-      getCountryOfOrigin(answers, recordId)
+      getCountryOfOrigin(answers, recordId, isCategorised)
     ).parMapN((eori, recordId, value) =>
       UpdateGoodsRecord(
         eori,
         recordId,
-        countryOfOrigin = Some(value),
-        category = Some(1)
+        countryOfOrigin = Some(value)
```
ChrisCookOC, line 53: Don't we need to set the category to None? Or is that a later thing?

app/models/UpdateGoodsRecord.scala
```diff
@@ -81,21 +82,38 @@ object UpdateGoodsRecord {
...
-        }
-      case Right(false) => Left(NonEmptyChain.one(UnexpectedPage(HasCommodityCodeChangePage(recordId))))
-      case Left(errors) => Left(errors)
+  private def getCommodityCode(
+    answers: UserAnswers,
+    recordId: String,
+    isCategorised: Boolean
+  ): EitherNec[ValidationError, Commodity] =
+    if (isCategorised) {
+      answers.getPageValue(HasCommodityCodeChangePage(recordId)) match {
```
ChrisCookOC: Could be worth sticking the cases in a function just to avoid people's brains melting looking at this function :D

app/models/UpdateGoodsRecord.scala
```diff
@@ -81,21 +82,38 @@ object UpdateGoodsRecord {
...
+                  Left(NonEmptyChain.one(UnexpectedPage(HasCorrectGoodsCommodityCodeUpdatePage(recordId))))
+                case Left(errors) => Left(errors)
+              }
+            case Left(errors) => Left(errors)
+          }
+        case Right(false) => Left(NonEmptyChain.one(UnexpectedPage(HasCommodityCodeChangePage(recordId))))
+        case Left(errors) => Left(errors)
+      }
+    } else {
+      answers.getPageValue(CommodityCodeUpdatePage(recordId)) match {
```
ChrisCookOC: Alternatively this is the same as the case Right(true) case above so could just have this logic in a function and reuse it?

app/viewmodels/checkAnswers/CommodityCodeSummary.scala
```diff
@@ -40,9 +40,17 @@ object CommodityCodeSummary {
...
+  ): SummaryListRow = {
     val changeLink = mode match {
-      case NormalMode => routes.HasCommodityCodeChangeController.onPageLoad(mode, recordId).url
+      case NormalMode =>
+        if (recordIsCategorised) {
+          routes.HasCommodityCodeChangeController.onPageLoad(mode, recordId).url
+        } else {
+          routes.CommodityCodeController.onPageLoadUpdate(mode, recordId).url
+        }
+      //routes.HasCommodityCodeChangeController.onPageLoad(mode, recordId).url
```
ChrisCookOC: Commented out code

yohan-oc: Updated

app/models/UpdateGoodsRecord.scala
Ewan-James-Donovan: +1

app/models/UpdateGoodsRecord.scala
yohan-oc: Updated

app/models/UpdateGoodsRecord.scala
yohan-oc: Updated :)

app/models/UpdateGoodsRecord.scala
yohan-oc: This is default set to `None` https://github.com/hmrc/trader-goods-profiles-frontend/pull/332/files#diff-0d95f1a32505dd8cdc88e6597445c3bcebc79eaf8d2832c9cb02b26466b5cd60L32

app/controllers/CyaUpdateRecordController.scala
yohan-oc: This is now handled with `recoverWith`

app/controllers/CyaUpdateRecordController.scala
```diff
@@ -209,27 +266,42 @@ class CyaUpdateRecordController @Inject() (
...
+              )
+          }
+        }
+        .recoverWith { case e: Exception =>
+          logger.error(s"Unable to fetch record $recordId: ${e.getMessage}")
           Future.successful(
-            logErrorsAndContinue(errors, routes.CyaUpdateRecordController.onPageLoadCommodityCode(recordId))
+            Redirect(
+              routes.JourneyRecoveryController
+                .onPageLoad(Some(RedirectUrl(routes.CyaUpdateRecordController.onPageLoadCommodityCode(recordId).url)))
```
ChrisCookOC: Probably shouldn't redirect back to the page we are already on - we know it doesn't work! 
Should probably go to the good record? Or the start of the journey

yohan-oc: Ops. You are right. Update the continue url to (Single) Goods record page

ChrisCookOC (dismissed):

ChrisCookOC (approved):

mycp98 (approved):

vinoocs (approved):

PR #335, opened 2024-08-12
TGP 2066 : record lock advice status

vinoocs:


Ewan-James-Donovan (approved):

app/controllers/SingleRecordController.scala
```diff
@@ -52,6 +52,14 @@ class SingleRecordController @Inject() (
     (identify andThen profileAuth andThen getData andThen requireData).async { implicit request =>
       for {
         record                             <- goodsRecordConnector.getRecord(request.eori, recordId)
+        recordIsLocked                      = record.adviceStatus match {
```
ChrisCookOC, line 55: Not necessary if you are happy, but if you want to, you could add this as a method on the record itself, and then just call that here

test/controllers/SingleRecordControllerSpec.scala
```diff
@@ -239,5 +248,298 @@ class SingleRecordControllerSpec extends SpecBase with MockitoSugar {
       }
     }
 
+    "TraderReferenceSummary.row" - {
```
ChrisCookOC, line 251: These tests are in the wrong place?

ChrisCookOC (approved):

test/controllers/SingleRecordControllerSpec.scala
vinoocs: No, they are in the right place. As we display these rows on single record page, added these tests here.

mycp98 (approved):

app/controllers/SingleRecordController.scala
vinoocs: can do this as improvement in the other PR.

PR #340, opened 2024-08-12
TGP-2035 - refactor categorisation

ChrisCookOC:


conf/application.conf
```diff
@@ -130,6 +130,10 @@ enrolment-config {
 # Default value for local environment
 play.frontend.host = "http://localhost:10905"
 
+auditing {
```
besscerule: should this be included?

conf/app.routes
```diff
@@ -147,17 +147,26 @@ POST        /update-record/:recordId/categorisation/category-assessment/:index
...
+GET         /update-record/:recordId/categorisation/result/standard                         controllers.CategorisationResultController.onPageLoad(recordId: String, scenario: Scenario = StandardGoodsScenario)
+GET         /update-record/:recordId/categorisation/result/category-1-no-exemptions         controllers.CategorisationResultController.onPageLoad(recordId: String, scenario: Scenario = Category1NoExemptionsScenario)
+GET         /update-record/:recordId/categorisation/result/standard-no-assessments          controllers.CategorisationResultController.onPageLoad(recordId: String, scenario: Scenario = StandardGoodsNoAssessmentsScenario)
 
 GET         /update-record/:recordId/categorisation/check-your-answers                      controllers.CyaCategorisationController.onPageLoad(recordId:String)
 POST        /update-record/:recordId/categorisation/check-your-answers                      controllers.CyaCategorisationController.onSubmit(recordId:String)
 
+GET         /update-record/:recordId/categorisation/categorisationPreparation               controllers.CategorisationPreparationController.startCategorisation(recordId:String)
```
besscerule: these paths should have hyphens like

`/update-record/:recordId/categorisation/categorisation-preparation`

app/viewmodels/checkAnswers/AssessmentsSummary.scala
```diff
@@ -18,30 +18,44 @@ package viewmodels.checkAnswers
...
 
   def row(
     recordId: String,
     answers: UserAnswers,
     assessment: CategoryAssessment,
-    indexOfThisAssessment: Int
-  )(implicit messages: Messages): Option[SummaryListRow] =
-    answers.get(AssessmentPage(recordId, indexOfThisAssessment)).map { answer =>
+    indexOfThisAssessment: Int,
+    reassessmentAnswer: Boolean
```
besscerule: given this is a boolean flag maybe we could change is to isReassementAnswer - or something that makes more sense?

app/viewmodels/checkAnswers/AssessmentsSummary.scala
```diff
@@ -18,30 +18,44 @@ package viewmodels.checkAnswers
...
+  )(implicit messages: Messages): Option[SummaryListRow] = {
+
+    val pageToUse = if (reassessmentAnswer) {
+      ReassessmentPage(recordId, indexOfThisAssessment)
+    } else {
+      AssessmentPage(recordId, indexOfThisAssessment)
+    }
+
+    val changeLink = if (reassessmentAnswer) {
```
besscerule: this feels a bit non-functional with two if statements assessing the same thing, pretty sure you could shoehorn most of this into line 50 so you would only need to check once

app/viewmodels/checkAnswers/CategorySummary.scala
```diff
@@ -53,7 +55,7 @@ object CategorySummary {
         } else {
           ValueViewModel(
             HtmlContent(
-              s"<a href=${routes.CategoryGuidanceController.onPageLoad(recordId).url} class='govuk-link'>$translatedValue</a>"
+              s"<a href=${routes.CategorisationPreparationController.startCategorisation(recordId).url} class='govuk-link'>$translatedValue</a>"
```
besscerule: you use this url twice so have it in a variable

app/viewmodels/checkAnswers/HasSupplementaryUnitSummary.scala
```diff
@@ -88,4 +88,5 @@ object HasSupplementaryUnitSummary {
     } else {
       None
     }
+
```
besscerule: remove this

app/viewmodels/checkAnswers/LongerCommodityCodeSummary.scala
```diff
@@ -17,41 +17,30 @@
...
-          SummaryListRowViewModel(
-            key = "longerCommodityCode.checkYourAnswersLabel",
-            value = ValueViewModel(info.commodityCode),
-            actions = Seq(
-              ActionItemViewModel(
-                "site.change",
-                routes.LongerCommodityCodeController.onPageLoad(CheckMode, recordId).url
-              )
-                .withVisuallyHiddenText(messages("longerCommodityCode.change.hidden"))
+  def row(answers: UserAnswers, recordId: String)(implicit messages: Messages): Option[SummaryListRow] =
```
besscerule, line 29: this is way nicer! :)

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
```diff
@@ -91,4 +90,5 @@ object SupplementaryUnitSummary {
         }
       )
     }
+
```
besscerule: remove this line

app/services/CategorisationService.scala
```diff
@@ -16,216 +16,124 @@
...
-        }
+  def getCategorisationInfo(
+    request: DataRequest[_],
+    commodityCode: String,
+    country: String,
+    recordId: String,
+    longerCode: Boolean = false
+  )(implicit hc: HeaderCarrier): Future[CategorisationInfo] = {
+
+    val ottResponse = ottConnector.getCategorisationInfo(
```
besscerule: you don't need to have this as a separate variable just flat map straight from it

besscerule: also - does this even need to be in a service? it feels quite a small thing it is doing and could be directly in the controller maybe?

app/services/CategorisationService.scala
```diff
@@ -16,216 +16,124 @@
...
+      Some(recordId),
+      country,
+      LocalDate.now()
+    )
+
+    ottResponse.flatMap { response =>
+      CategorisationInfo.build(response, commodityCode, longerCode) match {
+        case Some(categorisationInfo) => Future.successful(categorisationInfo)
+        case _                        =>
+          Future.failed(new RuntimeException("Could not build categorisation info"))
```
besscerule: shouldn't this redirect to a JR page?

app/services/CategorisationService.scala
```diff
@@ -16,216 +16,124 @@
...
-            recordCategorisations.copy(records = recordCategorisations.records + (recordId -> categorisationInfo))
-          )
-        )
-      updatedAnswersCleanUpAssessments <-
-        Future.fromTry(cleanupOldAssessmentAnswers(updatedAnswers, recordId))
-      _                                <- sessionRepository.set(updatedAnswersCleanUpAssessments)
-    } yield updatedAnswersCleanUpAssessments
-  }
+  ): Scenario =
+    if (categorisationInfo.categoryAssessments.isEmpty) {
```
besscerule, line 64: i'm just assuming that you got this right here - i don't think you can simply the ifs anymore

app/services/CategorisationService.scala
```diff
@@ -16,216 +16,124 @@
...
-      count               = categorisationInfo.categoryAssessments.size
-      //Go backwards to avoid recursion issues
-      rangeToRemove       = (firstAssessmentIndex to count + 1).reverse
-    } yield rangeToRemove.foldLeft[Try[UserAnswers]](Success(userAnswers)) { (acc, currentIndexToRemove) =>
-      acc.flatMap(_.remove(AssessmentPage(recordId, currentIndexToRemove)))
-    }).getOrElse(
-      Failure(new InconsistentUserAnswersException(s"Could not find category assessments"))
-    )
+  ) = {
+    val listOfAnswers = categorisationInfo.getAnswersForQuestions(userAnswers, recordId)
```
besscerule: do you need to have these as separate variables?

app/services/CategorisationService.scala
```diff
@@ -16,216 +16,124 @@
...
+        }
+    }
 
-      val cleanedUserAnswers = cleanupOldAssessmentAnswers(userAnswers, recordId).get
-      // Avoid it getting upset if answers have moved too far
-      val uaWithPlaceholders = newAssessments.zipWithIndex.foldLeft[Try[UserAnswers]](Success(cleanedUserAnswers)) {
-        (currentAnswers, newAssessment) =>
-          currentAnswers.flatMap(_.set(AssessmentPage(recordId, newAssessment._2), NotAnsweredYet))
-      }
+    // Avoid it getting upset if answers have moved too far
```
besscerule, line 117: this made me giggle

besscerule (commented):
some comments

app/viewmodels/checkAnswers/CategorySummary.scala
yohan-oc: Fixed :)

app/viewmodels/checkAnswers/HasSupplementaryUnitSummary.scala
yohan-oc: Removed :)

app/services/CategorisationService.scala
Ewan-James-Donovan: I think since this service is used in multiple places, it's a big enough thing to not repeat in two places

app/services/CategorisationService.scala
ChrisCookOC: The return type doesn't allow that, so the place that calls it needs to handle that.

Could change the return type but that complicates logic more I think?

app/services/CategorisationService.scala
Ewan-James-Donovan: Could kind of 'flatten' it with pattern matching if you think it's worth it: (DON'T ASSUME THIS CODE IS CORRECT)
```
categorisationInfo match {
  case info if info.categoryAssessments.isEmpty =>
    StandardGoodsNoAssessmentsScenario

  case info 
    if info.categoryAssessmentsThatNeedAnswers.isEmpty && 
       info.categoryAssessments.exists(_.isCategory1) =>
    Category1NoExemptionsScenario

  case info if info.categoryAssessmentsThatNeedAnswers.isEmpty =>
    Category2Scenario

  case _ =>
    calculateBasedOnAnswers(categorisationInfo, userAnswers, recordId)
}
```

conf/application.conf
ChrisCookOC: Definitely not 🙈 good spot :)

app/services/CategorisationService.scala
Ewan-James-Donovan: I'm not sure that's any nicer now I look at it actually :')

Ewan-James-Donovan:
I'm very happy with this. The simplifications brought in and the separation of concerns is making my dev heart smile.

app/services/CategorisationService.scala
ChrisCookOC: That could work. Would hold off for now just because its one of two places that need to change for Niphls/Nirms which will mean it changes soon anyway :)

app/controllers/CategorisationResultController.scala
```diff
@@ -18,35 +18,28 @@ package controllers
...
   override val messagesApi: MessagesApi,
   identify: IdentifierAction,
   getData: DataRetrievalAction,
   requireData: DataRequiredAction,
   val controllerComponents: MessagesControllerComponents,
-  view: CategorisationResultView,
-  sessionRepository: SessionRepository
-)(implicit ec: ExecutionContext)
+  view: CategorisationResultView
+)(implicit @unused ec: ExecutionContext)
```
liandra-giacomine: why do we need it if it's unused?

app/controllers/SupplementaryUnitController.scala
```diff
@@ -61,34 +61,37 @@ class SupplementaryUnitController @Inject() (
...
-              Future.successful(BadRequest(view(formWithErrors, mode, recordId, measurementUnit, onSubmitAction)))
+            val catInfo = request.userAnswers.get(LongerCategorisationDetailsQuery(recordId)) match {
+              case Some(catInfo) => Some(catInfo)
+              case _             => request.userAnswers.get(CategorisationDetailsQuery(recordId))
             }
-            result.getOrElse(Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url)))
+
+            catInfo
+              .map { categorisationInfo =>
+                val measurementUnit = categorisationInfo.measurementUnit.getOrElse("")
```
liandra-giacomine: when this is "allowed" to be empty the option should be None, why are we doing getOrElse("") here? :)

ChrisCookOC: It seems to be driven by the view. Tempted to change it. I think if the view should decide what it displays rather than the controller having to care :)

app/controllers/CategorisationResultController.scala
ChrisCookOC: So yeah on this one that is right we can remove it

On some of the others, one of the functions is using it so its showing as unused in the class for some reason, so those ones need to be flagged unused to keep the compiler happy :)

conf/app.routes
ChrisCookOC: Fixed :)

app/viewmodels/checkAnswers/AssessmentsSummary.scala
ChrisCookOC: Fixed :)

app/viewmodels/checkAnswers/AssessmentsSummary.scala
ChrisCookOC: Combined the two statements for now, how does that look?

app/viewmodels/checkAnswers/SupplementaryUnitSummary.scala
ChrisCookOC: Done :) , another casualty in Beth's war on the empty line 😆

app/services/CategorisationService.scala
ChrisCookOC: Cleaned up a bit :)

app/controllers/SupplementaryUnitController.scala
ChrisCookOC: Changed this now so the view deals with the option :) Also means its in one place rather than 4!

liandra-giacomine (changes requested):
I think we can update a few of these if statements to be more clean

app/controllers/CategorisationResultController.scala
```diff
@@ -18,35 +18,25 @@ package controllers
...
-  view: CategorisationResultView,
-  sessionRepository: SessionRepository
-)(implicit ec: ExecutionContext)
-    extends FrontendBaseController
+  view: CategorisationResultView
+) extends FrontendBaseController
     with I18nSupport {
 
   def onPageLoad(recordId: String, scenario: Scenario): Action[AnyContent] =
     (identify andThen getData andThen requireData) { implicit request =>
```
vinoocs: can be accessed without a profile and user can click on the links  please can we add profileAuth like in other controllers

app/controllers/CyaCategorisationController.scala
```diff
@@ -50,102 +50,126 @@ class CyaCategorisationController @Inject() (
...
+    errorMessage: String,
+    recordId: String,
+    userAnswers: UserAnswers
+  ): Result = {
 
-    logger.error(s"Unable to update Goods Profile.  Missing pages: $errorMessages")
-    dataCleansingService.deleteMongoData(request.userAnswers.id, CategorisationJourney)
+    val continueUrl = RedirectUrl(routes.CategorisationPreparationController.startCategorisation(recordId).url)
+
+    logger.error(s"Unable to update Goods Profile: $errorMessage")
```
vinoocs: Please can we add recordId for the log like we have in other places

app/controllers/HasSupplementaryUnitController.scala
```diff
@@ -51,13 +50,8 @@ class HasSupplementaryUnitController @Inject() (
...
   private val form = formProvider()
 
-  def onPageLoad(mode: Mode, recordId: String): Action[AnyContent] =
-    (identify andThen profileAuth andThen getData andThen requireData) { implicit request =>
-      for {
-        updatedUA <- Future.fromTry(request.userAnswers.set(RecategorisingQuery(recordId), false))
-        _         <- sessionRepository.set(updatedUA)
-      } yield updatedUA
-
+  def onPageLoad(mode: Mode, recordId: String): Action[AnyContent] = (identify andThen getData andThen requireData) {
```
vinoocs: we removed profileAuth?

ChrisCookOC: Oops good spot, added it back!

app/controllers/CyaCategorisationController.scala
ChrisCookOC: Done :)

app/controllers/CategorisationResultController.scala
ChrisCookOC: Good shout, added :)

liandra-giacomine (dismissed):

vinoocs (approved):

PR #352, opened 2024-08-14
TGP-2086 : Commodity code expired day

vinoocs:
Moved to internally blocked due to conflicting changes in https://jira.tools.tax.service.gov.uk/browse/TGP-2035 .. Also update being PATCH , we should always retrieve commodity validity from OTT & not rely on record

### DIFF
--- a/app/controllers/CategoryGuidanceController.scala
+++ b/app/controllers/CategoryGuidanceController.scala
@@ -31,6 +31,8 @@ import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import utils.SessionData.{dataRemoved, dataUpdated, pageUpdated}
 import views.html.CategoryGuidanceView
 
+import java.time.temporal.ChronoUnit
+import java.time.{Instant, ZoneId, ZonedDateTime}
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -44,8 +46,8 @@ class CategoryGuidanceController @Inject() (
   view: CategoryGuidanceView,
   auditService: AuditService,
   categorisationService: CategorisationService,
-  navigator: Navigator,
-  goodsRecordConnector: GoodsRecordConnector
+  goodsRecordsConnector: GoodsRecordConnector,
+  navigator: Navigator
 )(implicit ec: ExecutionContext)
     extends FrontendBaseController
     with I18nSupport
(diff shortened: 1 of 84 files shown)


conf/app.routes
```diff
@@ -265,9 +265,12 @@ GET        /update-record/:recordId/withdraw-request/success
...
 GET        /update-record/:recordId/withdraw-request/reason                                 controllers.ReasonForWithdrawAdviceController.onPageLoad(recordId: String)
 POST       /update-record/:recordId/withdraw-request/reason                                 controllers.ReasonForWithdrawAdviceController.onSubmit(recordId: String)
+GET        /problem/expired-commodity-code                                                  controllers.ExpiredCommodityCodeController.onPageLoad(recordId:String)
 
 GET        /download-data/request-file                                                      controllers.RequestDataController.onPageLoad()
 POST       /download-data/request-file                                                      controllers.RequestDataController.onSubmit(email: String)
 
 # this url must be the last url on this page so it becomes the default if no other urls match
 GET        /*path                                                                           controllers.PageNotFoundController.onPageLoad(path: String)
+
```
besscerule, line 281: can you remove this extra line

conf/app.routes
```diff
@@ -265,9 +265,12 @@ GET        /update-record/:recordId/withdraw-request/success
 
 GET        /update-record/:recordId/withdraw-request/reason                                 controllers.ReasonForWithdrawAdviceController.onPageLoad(recordId: String)
 POST       /update-record/:recordId/withdraw-request/reason                                 controllers.ReasonForWithdrawAdviceController.onSubmit(recordId: String)
+GET        /problem/expired-commodity-code                                                  controllers.ExpiredCommodityCodeController.onPageLoad(recordId:String)
```
besscerule: i think you need to have a record id in this url because otherwise if i go straight to that url (type it into the address bar) it says bad request because the record id will be null

app/controllers/SingleRecordController.scala
```diff
@@ -45,7 +46,7 @@ class SingleRecordController @Inject() (
   dataCleansingService: DataCleansingService,
   val controllerComponents: MessagesControllerComponents,
   view: SingleRecordView
-)(implicit ec: ExecutionContext)
+)(implicit @unused ec: ExecutionContext)
```
besscerule, line 49: can you just remove this if it is unused?

conf/app.routes
vinoocs: done it!

conf/app.routes
vinoocs: I believe this should be navigated through previous page, url hacking isn't fair. Also, this url came from design, and recordId acts as a query param here so when user navigates from previous it gets appended.

app/controllers/SingleRecordController.scala
vinoocs: yeah, something that I asked in the chat yesterday, we can't remove implicit ec - it is needed. But for some reasons, it shows the warning so we're suppressing it. Please let me know if there's a way around to fix this without removing

conf/app.routes
besscerule: i think we should push back and get the recordId in the url - i'm sure if you explain to design they would be happy for the recordId to be in it

besscerule: because if not we have urls that are broken, which just feels like the wrong way to build an app

vinoocs: yeah, sorry didn't realise about potential issues - it's updated now please.

app/controllers/CategoryGuidanceController.scala
```diff
@@ -44,40 +46,49 @@ class CategoryGuidanceController @Inject() (
...
-                    .updateCategoryAndComcodeForGoodsRecord(request.eori, recordId, categoryRecord)
-                    .map { _ =>
-                      Redirect(routes.CategorisationResultController.onPageLoad(recordId, scenario.get).url)
-                        .removingFromSession(dataUpdated, pageUpdated, dataRemoved)
-                    }
+      // Check if the commodity code is expired
+      goodsRecordsConnector
+        .getRecord(eori = request.eori, recordId = recordId)
+        .flatMap { record =>
+          if (isCommCodeExpired(record.comcodeEffectiveToDate)) {
```
ChrisCookOC, line 62: Just a note this will have to move when merged into categorisation
(I'm assuming you'll merge first :) )

app/controllers/CategoryGuidanceController.scala
```diff
@@ -44,40 +46,49 @@ class CategoryGuidanceController @Inject() (
...
+            Future.successful(Redirect(routes.ExpiredCommodityCodeController.onPageLoad(recordId).url))
```
ChrisCookOC, line 63: Again after merging with categorisation, this needs to happen in the navigator.

app/controllers/CyaUpdateRecordController.scala
```diff
@@ -128,8 +129,14 @@ class CyaUpdateRecordController @Inject() (
       goodsRecordConnector
         .getRecord(request.eori, recordId)
         .flatMap { recordResponse =>
+          val navigatingFromExpiredCommCode = request.session.get(fromExpiredCommodityCodePage).contains("true")
```
ChrisCookOC, line 132: Is this necessarily needed? 

You have the goods record already so you can tell if its expired from that, rather than the page you've been :)

app/models/UpdateGoodsRecord.scala
```diff
@@ -67,11 +71,12 @@ object UpdateGoodsRecord {
   def validateCommodityCode(
     answers: UserAnswers,
     recordId: String,
-    isCategorised: Boolean
+    isCategorised: Boolean,
+    navigatingFromExpiredCommCode: Boolean
```
ChrisCookOC: I'd rename it to be "isComCodeExpired" or something rather than "navigatingFromExpired". Slight difference but it makes clear why we are doing this.

app/controllers/CyaUpdateRecordController.scala
vinoocs: yeah, I had that code earlier but due to more duplications, I wanted to keep it simple like this with a flag.

app/viewmodels/checkAnswers/CategorySummary.scala
```diff
@@ -62,4 +79,24 @@ object CategorySummary {
         value = viewModel
       )
     }
+  }
+
+  private def isCommCodeExpired(commcodeEffectiveToDate: Option[Instant]): Boolean = {
```
ChrisCookOC: As its in multiple places we can probably put it more central. A function on the goods record perhaps?

Ewan-James-Donovan (approved):

ChrisCookOC:
Lots of conflicts with the refactor PR - we probably want to merge that first then readjust this / rebranch off that

PR #358, opened 2024-08-16
TGP-1669 :audit event for withdraw advice

vinoocs:


app/services/AuditService.scala
```diff
@@ -133,6 +135,24 @@ class AuditService @Inject() (auditConnector: AuditConnector, auditEventFactory:
...
   }
 
+  def auditWithdrawAdvice(affinityGroup: AffinityGroup, eori: String, recordId: String, withdrawReason: Option[String])(
+    implicit hc: HeaderCarrier
+  ): Future[Done] = {
+    val event =
+      auditEventFactory.createWithdrawAdviceEvent(
+        affinityGroup,
+        eori,
+        RequestAdviceJourney,
```
Ewan-James-Donovan: Called about this, maybe needs to be WithdrawRequestJourney

vinoocs: Thanks for spotting that @Ewan-James-Donovan, updated now!

Ewan-James-Donovan (approved):

PR #364, opened 2024-08-19
TGP-2248: Added a hidden label for screen readers.

Ewan-James-Donovan:
The issue was that the select element that the autocomplete JS thing hides needed a label. Since it's only hidden, not deleted, it does need a label. It can't be deleted either, as the accessible autocomplete relies on it.

I added a hidden label, and used a screen reader to check it all behaves the same :)

### DIFF
--- a/app/assets/javascripts/app.js
+++ b/app/assets/javascripts/app.js
@@ -60,9 +60,6 @@ if (typeof accessibleAutocomplete != 'undefined' && document.querySelector('.aut
     selectEl.parentNode.insertBefore(label, selectEl);
   }
 
-
-
-
   // =====================================================
   // Polyfill autocomplete once loaded
   // =====================================================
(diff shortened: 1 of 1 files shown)


ChrisCookOC (dismissed):

app/assets/javascripts/app.js
```diff
@@ -47,8 +52,17 @@ if (typeof accessibleAutocomplete != 'undefined' && document.querySelector('.aut
...
     });
+
+    // Overwrite 'for' attribute of the hidden label after autocomplete changes are applied
+    label.setAttribute("for", document.querySelector('.autocomplete').id);
+
+    // Insert the hidden label immediately before the selectEl
+    selectEl.parentNode.insertBefore(label, selectEl);
   }
 
+
```
besscerule: Can we get rid of these spare lines

besscerule (dismissed):

app/assets/javascripts/app.js
Ewan-James-Donovan: Done @besscerule :)

ChrisCookOC (approved):

ChrisCookOC:
Like the comments explaining why we are having to do a seemingly odd thing :)

PR #365, opened 2024-08-20
TGP-1971: Implement Download data link behaviour

besscerule:
…ods profile


app/controllers/GoodsRecordsController.scala
```diff
@@ -100,6 +100,29 @@ class GoodsRecordsController @Inject() (
...
 
+  def getDownloadLinkMessagesKey(opt: Option[DownloadDataSummary]): String =
+    opt match {
+      case Some(downloadDataSummary) if downloadDataSummary.status == RequestFile    =>
+        "goodsRecord.downloadLinkText.requestFile"
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileInProgress =>
+        "goodsRecord.downloadLinkText.fileInProgress"
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileReady      =>
+        "goodsRecord.downloadLinkText.fileReady"
+      case _                                                                         =>
```
vinoocs, line 111: do we cover these cases as part of tests

app/controllers/GoodsRecordsController.scala
```diff
@@ -100,6 +100,29 @@ class GoodsRecordsController @Inject() (
...
+    }
+
+  def getDownloadLinkRoute(opt: Option[DownloadDataSummary]): String =
+    opt match {
+      case Some(downloadDataSummary) if downloadDataSummary.status == RequestFile    =>
+        routes.RequestDataController.onPageLoad().url
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileInProgress =>
+        routes.FileInProgressController.onPageLoad().url
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileReady      =>
+        routes.FileReadyController.onPageLoad().url
```
vinoocs, line 122: please can we have test cases for these

app/models/DownloadDataSummary.scala
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+import play.api.libs.json.{Format, JsError, JsResult, JsString, JsSuccess, JsValue, Json, OFormat}
+
+sealed abstract class DownloadDataStatus(val value: String)
+
+object DownloadDataStatus {
```
vinoocs: may be worth moving it into its own file than having it here?

app/controllers/GoodsRecordsController.scala
besscerule: good spot!

app/models/DownloadDataSummary.scala
besscerule: Have looked at some other repos on hmrc and this seems to be the pattern since they are both closely related - what do you think?

vinoocs: right, just pedantic. Looks like we have the same pattern in multiple places :)

app/connectors/DownloadDataConnector.scala
```diff
@@ -0,0 +1,65 @@
...
+class DownloadDataConnector @Inject() (config: Configuration, httpClient: HttpClientV2)(implicit
+  ec: ExecutionContext
+) {
+  private val dataStoreBaseUrl: Service = config.get[Service]("microservice.services.trader-goods-profiles-data-store")
+  private val clientIdHeader            = ("X-Client-ID", "tgp-frontend")
+
+  private def downloadDataSummaryUrl(eori: String) =
+    url"$dataStoreBaseUrl/trader-goods-profiles-data-store/traders/$eori/download-data-summary"
+
+  def requestDownloadData(eori: String)(implicit hc: HeaderCarrier): Future[Boolean] =
```
ChrisCookOC, line 38: This is only used in tests right now. Will we need it later?

besscerule: yes

vinoocs (dismissed):

vinoocs (approved):

PR #366, opened 2024-08-21
TGP-1399/base controller trait

Ewan-James-Donovan:
This is a base controller trait which we can update as we see fit. Currently it only handles logErrorsAndContinue as this was the only consistently required thing I could see.

### DIFF
--- a/app/controllers/AdviceStartController.scala
+++ b/app/controllers/AdviceStartController.scala
@@ -20,9 +20,8 @@ import controllers.actions._
 import models.NormalMode
 import navigation.Navigator
 import pages.AdviceStartPage
-import play.api.i18n.{I18nSupport, MessagesApi}
+import play.api.i18n.MessagesApi
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
-import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import utils.SessionData.{dataRemoved, dataUpdated, pageUpdated}
 import views.html.AdviceStartView
 
@@ -37,9 +36,7 @@ class AdviceStartController @Inject() (
   val controllerComponents: MessagesControllerComponents,
   view: AdviceStartView,
   navigator: Navigator
-) extends FrontendBaseController
-    with I18nSupport
-    with BaseController {
+) extends BaseController {
 
   def onPageLoad(recordId: String): Action[AnyContent] =
     (identify andThen profileAuth andThen getData andThen requireData) { implicit request =>
--- a/app/controllers/AdviceStartController.scala
+++ b/app/controllers/AdviceStartController.scala
@@ -38,7 +38,8 @@ class AdviceStartController @Inject() (
   view: AdviceStartView,
   navigator: Navigator
 ) extends FrontendBaseController
-    with I18nSupport {
+    with I18nSupport
+    with BaseController {
 
   def onPageLoad(recordId: String): Action[AnyContent] =
     (identify andThen profileAuth andThen getData andThen requireData) { implicit request =>
--- a/app/controllers/BaseController.scala
+++ b/app/controllers/BaseController.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package controllers
 
 import cats.data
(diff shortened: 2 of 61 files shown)


app/controllers/AdviceStartController.scala
```diff
@@ -38,7 +38,8 @@ class AdviceStartController @Inject() (
   view: AdviceStartView,
   navigator: Navigator
 ) extends FrontendBaseController
-    with I18nSupport {
+    with I18nSupport
```
ChrisCookOC: Since every controller extends FrontendBaseController with I18nSupport with BaseController, can we just make the BaseController extends FrontendBaseController with I18nSupport, and then everyone of our controllers just extends BaseController?

Ewan-James-Donovan: I did consider that. Since you're saying this now, I'll try to do it :)

Ewan-James-Donovan: Done @ChrisCookOC

app/controllers/BaseController.scala
```diff
@@ -0,0 +1,62 @@
...
+import models.requests.DataRequest
+import play.api.i18n.I18nSupport
+import play.api.mvc.{AnyContent, Call, Result}
+import services.DataCleansingService
+import uk.gov.hmrc.play.bootstrap.binders.RedirectUrl
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
+
+trait BaseController extends FrontendBaseController with I18nSupport with Logging {
+
+  @Inject var dataCleansingService: DataCleansingService = _
```
ChrisCookOC: What is this doing / it looks scary :D

Ewan-James-Donovan: It's a field injector, can't be 'val' because of the fact it needs to be injected. I can't see any way to inject directly to the trait other than this, trying the regular way doesn't seem to work.

Practically speaking, it saves having to inject on every class that requires it (it's used by one of the logErrorsAndContinue methods of the trait).

If you know of a better way, do tell!

ChrisCookOC: Tried a few things and failed :( 

Can we just pass in the dataCleansingService to that version of the logErrorsAndContinue? It's not used in many places as we usually don't

Idk, just something seems off to me about trying to get the BaseController to do it. Happy to be wrong though if other people know better :)

Ewan-James-Donovan: Currently occupied, but yeah sure we can just pass it in or something similar. I'll take a look tomorrow!

Ewan-James-Donovan: Made it implicit rather than injected. Seems like a good middle ground

app/controllers/BaseController.scala
```diff
@@ -0,0 +1,62 @@
...
+trait BaseController extends FrontendBaseController with I18nSupport with Logging {
+
+  def logErrorsAndContinue(
+    errorMessage: String,
+    continueUrl: Call,
+    errors: data.NonEmptyChain[ValidationError],
+    journeyToCleanse: Journey
+  )(implicit
+    request: DataRequest[AnyContent],
+    dataCleansingService: DataCleansingService
```
liandra-giacomine: why are we passing the DataCleansingService implicitly/should we be cleansing here or in the controller?

app/controllers/CyaCreateProfileController.scala
```diff
@@ -41,15 +36,16 @@ class CyaCreateProfileController @Inject() (
   getData: DataRetrievalAction,
   requireData: DataRequiredAction,
   checkProfile: ProfileCheckAction,
-  dataCleansingService: DataCleansingService,
   val controllerComponents: MessagesControllerComponents,
   view: CyaCreateProfileView,
   traderProfileConnector: TraderProfileConnector,
+  implicit val dataCleansingService: DataCleansingService,
```
liandra-giacomine: I don't see the advantage of making this implicit/I think we would benefit from doing the cleansing in the controller to make the functionality less hidden

app/controllers/BaseController.scala
liandra-giacomine: @Ewan-James-Donovan  @ChrisCookOC  In Scala 2 you won't be able to have parameters in a trait, though that is possible in Scala 3. You can use a class if you need that though. However in this case I believe the controller should be doing the cleansing and shouldn't be passed to the base controller as it hides functionality

app/controllers/BaseController.scala
Ewan-James-Donovan: I think the idea is that this data cleansing happens the same way for a fair few places, in the logErrorsAndContinue method which was present on a fair few controller, which this sought to move into one place. Implicitly passing in just saves explicitly passing it in, that's all

app/controllers/CyaCreateProfileController.scala
Ewan-James-Donovan: See above, pop me a message if you like, I'm concerned that by doing it in the controllers then we're just repeating behaviour

app/controllers/BaseController.scala
Ewan-James-Donovan: @liandra-giacomine I've done as you suggested :)

ChrisCookOC: yeah makes sense to have it separate :)

Ewan-James-Donovan: I've just fixed some bits :)

liandra-giacomine (approved):

PR #368, opened 2024-08-21
TGP-1388 - navigator do the navigatin'

ChrisCookOC:
skipped the categorisation ones as they are done in the refactor PR

skipped a couple of other cases where the navigation logic was too complicated to make changing as part of this ticket desirable


app/controllers/CyaUpdateRecordController.scala
```diff
@@ -200,10 +196,10 @@ class CyaUpdateRecordController @Inject() (
                               )
             updatedAnswers <- Future.fromTry(request.userAnswers.remove(TraderReferenceUpdatePage(recordId)))
             _              <- sessionRepository.set(updatedAnswers)
-          } yield Redirect(routes.SingleRecordController.onPageLoad(recordId))
+          } yield Redirect(navigator.nextPage(CyaUpdateRecordPage(recordId), NormalMode, updatedAnswers))
```
vinoocs, line 218: Pedantic. Can we just name it as UpdateRecordPage instead, it doesn't really act as a cya here I believe

app/controllers/CyaSupplementaryUnitController.scala
```diff
@@ -105,7 +108,7 @@ class CyaSupplementaryUnitController @Inject() (
 
           goodsRecordConnector.updateSupplementaryUnitForGoodsRecord(request.eori, recordId, model).map { _ =>
             dataCleansingService.deleteMongoData(request.userAnswers.id, SupplementaryUnitUpdateJourney)
-            Redirect(routes.SingleRecordController.onPageLoad(recordId))
+            Redirect(navigator.nextPage(CyaSupplementaryUnitPage(recordId), NormalMode, request.userAnswers))
```
vinoocs, line 96: can we name it as UpdateRecordPage instead and as it is not really a cya and we don't need an extra navigation in navigator

app/controllers/CyaUpdateRecordController.scala
liandra-giacomine: This raises a bigger question about why we have separated the update record into specific parts instead of reusing an update object

besscerule: cries in scala

besscerule (commented):
Can we get some tests for the missing code coverage in the navigator? The happy to approve - great work!

app/controllers/CyaUpdateRecordController.scala
ChrisCookOC: It does act as a CYA page here, e.g.
image

app/controllers/CyaSupplementaryUnitController.scala
ChrisCookOC: This is a CYA page

ChrisCookOC:
> Can we get some tests for the missing code coverage in the navigator? The happy to approve - great work!
@besscerule enjoy! 
image

Ewan-James-Donovan (approved):

PR #380, opened 2024-08-28
TGP-2043: Fix for back button behaviour on recategorisation 

yohan-oc:
**Fix for Recategorisation Answers Not Saving When Navigating Back and Forth Using the Back Button**

This is to ensures that when a user completes the recategorisation journey and reaches the CYA page, their recategorisation answers are saved correctly, even if they use the back button to navigate back and forth through the journey.

### DIFF
--- a/app/controllers/CategorisationPreparationController.scala
+++ b/app/controllers/CategorisationPreparationController.scala
@@ -19,7 +19,7 @@ package controllers
 import connectors.GoodsRecordConnector
 import controllers.actions.{DataRequiredAction, DataRetrievalAction, IdentifierAction, ProfileAuthenticateAction}
 import models.ott.CategorisationInfo
-import models.{CategoryRecord, Mode, NormalMode, UserAnswers}
+import models.{CategoryRecord, Commodity, Mode, NormalMode, UserAnswers}
 import navigation.Navigator
 import org.apache.pekko.Done
 import pages.{CategorisationPreparationPage, HasSupplementaryUnitPage, RecategorisationPreparationPage}
@@ -29,6 +29,7 @@ import queries.{CategorisationDetailsQuery, LongerCategorisationDetailsQuery, Lo
 import repositories.SessionRepository
 import services.CategorisationService
 import uk.gov.hmrc.http.HeaderCarrier
+import utils.Constants.minimumLengthOfCommodityCode
 
 import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
@@ -74,8 +75,9 @@ class CategorisationPreparationController @Inject() (
         shorterCategorisationInfo <-
           Future.fromTry(Try(request.userAnswers.get(CategorisationDetailsQuery(recordId)).get))
 
-        longerComCode            <- Future.fromTry(Try(request.userAnswers.get(LongerCommodityQuery(recordId)).get))
-        longerCategorisationInfo <-
+        oldLongerCategorisationInfoOpt = request.userAnswers.get(LongerCategorisationDetailsQuery(recordId))
+        longerComCode                 <- Future.fromTry(Try(request.userAnswers.get(LongerCommodityQuery(recordId)).get))
+        newLongerCategorisationInfo   <-
           categorisationService
             .getCategorisationInfo(
               request,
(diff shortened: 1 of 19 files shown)


app/controllers/CategorisationPreparationController.scala
```diff
@@ -89,22 +90,33 @@ class CategorisationPreparationController @Inject() (
...
-
-        updatedUAReassessmentAnswers <- Future.fromTry(
-                                          categorisationService.updatingAnswersForRecategorisation(
-                                            updatedUACatInfo,
-                                            recordId,
-                                            shorterCategorisationInfo,
-                                            longerCategorisationInfo
+          Future.fromTry(updatedUASuppUnit.set(LongerCategorisationDetailsQuery(recordId), newLongerCategorisationInfo))
+
+        isNewOneTheSameAsOldOne = oldLongerCategorisationInfoOpt.exists(_.equals(newLongerCategorisationInfo))
```
ChrisCookOC: Could we maybe put the logic from 104 to 117 in a function

then just call like 

updatedUAReassessmentAnswers <- newFunctionHere(updatedUACatInfo, recordId, shorterCatInfo, newLonger, oldLongerOpt)

test/controllers/CategorisationPreparationControllerSpec.scala
```diff
@@ -748,6 +748,62 @@ class CategorisationPreparationControllerSpec extends SpecBase with BeforeAndAft
 
       }
 
+      "and save the category information without copying answers from previous assessments" in {
```
ChrisCookOC: would be nice to mention the why here, e.g. 

"and save the category information without copying answers from previous assessments" - {

"when the longer commodity code was already set to the same value" in {

app/pages/ReassessmentPage.scala
```diff
@@ -25,15 +25,15 @@ import scala.util.{Failure, Success, Try}
...
   override def path: JsPath = JsPath \ "reassessments" \ recordId \ index
 
   override def cleanup(
-    value: Option[AssessmentAnswer],
+    value: Option[ReassessmentAnswer],
     updatedUserAnswers: UserAnswers,
     originalUserAnswers: UserAnswers
   ): Try[UserAnswers] =
-    if (value.contains(AssessmentAnswer.NoExemption)) {
+    if (value.exists(x => x.answer == AssessmentAnswer.NoExemption)) {
```
Ewan-James-Donovan: try not to use x as variable - I recall liandra saying something about this not being allowed in the final codebase

Ewan-James-Donovan (commented):
I like the use of Reassessment answer :) Just some small bits to consider

app/controllers/CategorisationPreparationController.scala
```diff
@@ -76,8 +76,9 @@ class CategorisationPreparationController @Inject() (
         shorterCategorisationInfo <-
           Future.fromTry(Try(request.userAnswers.get(CategorisationDetailsQuery(recordId)).get))
 
-        longerComCode            <- Future.fromTry(Try(request.userAnswers.get(LongerCommodityQuery(recordId)).get))
-        longerCategorisationInfo <-
+        oldLongerCategorisationInfoOpt = request.userAnswers.get(LongerCategorisationDetailsQuery(recordId))
+        longerComCode                 <- Future.fromTry(Try(request.userAnswers.get(LongerCommodityQuery(recordId)).get))
```
Ewan-James-Donovan, line 80: Assuming this is a '<-' future from try for a reason, instead of just '='

app/navigation/Navigator.scala
```diff
@@ -271,7 +271,7 @@ class Navigator @Inject() (categorisationService: CategorisationService) {
     answers.get(LongerCategorisationDetailsQuery(recordId)) match {
       case Some(catInfo) if catInfo.categoryAssessmentsThatNeedAnswers.nonEmpty =>
         val firstAnswer = answers.get(ReassessmentPage(recordId, firstAssessmentIndex))
-        if (answerIsEmpty(firstAnswer)) {
+        if (reassessmentAnswerIsEmpty(firstAnswer) || !firstAnswer.get.isAnswerCopiedFromPreviousAssessment) {
```
ChrisCookOC, line 281: Can we explicitly test this case where the first answer is not empty but also not copied so we want to go to the assessment controller (ditto checkmode below)

app/navigation/Navigator.scala
```diff
@@ -297,8 +297,11 @@ class Navigator @Inject() (categorisationService: CategorisationService) {
         assessmentQuestion <- categorisationInfo.getAssessmentFromIndex(assessmentPage.index)
         nextAnswer          = answers.get(ReassessmentPage(recordId, nextIndex))
         assessmentAnswer   <- answers.get(assessmentPage)
-      } yield assessmentAnswer match {
-        case AssessmentAnswer.Exemption if nextIndex < assessmentCount && answerIsEmpty(nextAnswer)              =>
+      } yield assessmentAnswer.answer match {
+        case AssessmentAnswer.Exemption
+            if nextIndex < assessmentCount && (reassessmentAnswerIsEmpty(nextAnswer) || !nextAnswer.exists(
```
ChrisCookOC, line 309: Can we explicitly test this case where the next answer is not empty but also not copied so we want to go to the assessment controller (also in check mode)

app/pages/ReassessmentPage.scala
yohan-oc: oh.. renamed it it now

app/controllers/CategorisationPreparationController.scala
ChrisCookOC: LongerCommodityQuery must exist as its set on the LongerCommodityCode HasCorrectGoods page, which is the step before this. If it doesn't exist, we can't process things here (well we could but would be a lot more hassle)

app/viewmodels/checkAnswers/AssessmentsSummary.scala
```diff
@@ -33,42 +33,53 @@ object AssessmentsSummary {
     assessment: CategoryAssessment,
     indexOfThisAssessment: Int,
     isReassessmentAnswer: Boolean
-  )(implicit messages: Messages): Option[SummaryListRow] = {
+  )(implicit messages: Messages): Option[SummaryListRow] =
+    if (isReassessmentAnswer) {
```
ChrisCookOC, line 61: A lot of duplicate code between the two cases. not too worried about that to be fair. If is an easy/nice/sensible way to reduce that great, otherwise don't worry.

app/controllers/CategorisationPreparationController.scala
yohan-oc: Yes, Chris told me if `request.userAnswers.get(LongerCommodityQuery(recordId)).get` fails `Try` will handle it safely

test/controllers/CategorisationPreparationControllerSpec.scala
yohan-oc: Updated

app/controllers/CategorisationPreparationController.scala
yohan-oc: I didn't see Chris's reply :D

app/controllers/CategorisationPreparationController.scala
yohan-oc: Refactored now

app/viewmodels/checkAnswers/AssessmentsSummary.scala
yohan-oc: Refactored

app/controllers/CategorisationPreparationController.scala
Ewan-James-Donovan: Cool :)

app/navigation/Navigator.scala
yohan-oc: Added a new test https://github.com/hmrc/trader-goods-profiles-frontend/pull/380/files#diff-d003f6f11c99e6244a53129b8487d5deaa4a15fdfc6ba66b531817382d9c2febR1590

app/navigation/Navigator.scala
yohan-oc: Added a new test https://github.com/hmrc/trader-goods-profiles-frontend/pull/380/files#diff-d003f6f11c99e6244a53129b8487d5deaa4a15fdfc6ba66b531817382d9c2febR3281

app/navigation/Navigator.scala
yohan-oc: Wait.. I think I need to add another test

yohan-oc: Required tests added now

ChrisCookOC (dismissed):

ChrisCookOC (approved):

PR #383, opened 2024-09-03
TGP-2385: updated nirms validation

besscerule:


app/forms/NirmsNumberFormProvider.scala
```diff
@@ -27,7 +26,24 @@ class NirmsNumberFormProvider @Inject() extends Mappings {
   def apply(): Form[String] =
     Form(
       "value" -> text("nirmsNumber.error.required")
-        .transform(removeWhitespace, identity[String])
+        .transform(removeHyphensAndSpaces, identity[String])
         .verifying(regexp(StringFieldRegex.nirmsRegex, "nirmsNumber.error.invalidFormat"))
+        .transform(addHyphens, identity[String])
     )
+
+  def removeHyphensAndSpaces: String => String = _.replaceAll(" ", "")
```
ChrisCookOC: does this actually remove hyphens anymore?

mauro-fanton: should this be still named removedWhiteSpace, as the function only remove spaces? Or it is the way round? the function should also remove hyphen (-)?

besscerule: good spot!

besscerule: nope - i changed it and i've changed the name too

app/forms/NirmsNumberFormProvider.scala
```diff
@@ -27,7 +26,24 @@ class NirmsNumberFormProvider @Inject() extends Mappings {
...
-        .transform(removeWhitespace, identity[String])
+        .transform(removeSpaces, identity[String])
         .verifying(regexp(StringFieldRegex.nirmsRegex, "nirmsNumber.error.invalidFormat"))
+        .transform(addHyphens, identity[String])
     )
+
+  def removeSpaces: String => String = _.replaceAll(" ", "")
+  def addHyphens: String => String = { original =>
+    val length = 11
+    if (original.length == length) {
```
mauro-fanton, line 37: just a tip, you may want to look into patch here: e.g. 

"abdet67867".patch(3, "-", 0).patch(5, "-", 0)
> val res3: String = abd-e-t67867

mauro-fanton (approved):

PR #384, opened 2024-09-03
TGP-2285: Bug fix for url manipulation error

yohan-oc:


app/controllers/AssessmentController.scala
```diff
@@ -162,8 +188,26 @@ class AssessmentController @Inject() (
...
+            handleDataCleansingAndRecovery(
+              request.userAnswers.id,
+              routes.CategorisationPreparationController.startLongerCategorisation(mode, recordId)
+            )
+          )
+        )
+        .recover(_ =>
+          handleDataCleansingAndRecovery(
+            request.userAnswers.id,
+            routes.CategorisationPreparationController.startLongerCategorisation(mode, recordId)
```
ChrisCookOC: Should not go here. Should go to normal startCategorisation. You will delete the data for longer categorisation when you do the data cleansing for Categorisation Journey

ChrisCookOC (changes requested):

app/controllers/AssessmentController.scala
yohan-oc: in both onPageLoadReassessment and onSubmitReassessment right?

ChrisCookOC: Yeah whenever we journey recovery :)

yohan-oc: Updated

ChrisCookOC (approved):

PR #390, opened 2024-09-04
TGP-2296 updated advice status tag being only Advice instead of Advic…

AmarkumarJagtap:
…e received


conf/messages.en
```diff
@@ -443,7 +443,7 @@ singleRecord.h2.categorisation = Categorisation
 singleRecord.category.row = Category
 singleRecord.h2.supplementaryUnit = Supplementary Unit
 singleRecord.h2.adviceStatus = Advice status
-singleRecord.adviceStatus.row = Advice received
```
besscerule, line 446: do we need welsh too?

AmarkumarJagtap: welsh message will be added as part of separate ticket, did discuss with Amaka on this

ChrisCookOC (approved):

PR #402, opened 2024-09-10
TGP-2303: implement-file-download-and-retention 

besscerule:


Ewan-James-Donovan (dismissed):
LGTM

app/controllers/GoodsRecordsController.scala
```diff
@@ -102,25 +102,29 @@ class GoodsRecordsController @Inject() (
...
-      case Some(downloadDataSummary) if downloadDataSummary.status == FileInProgress =>
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileInProgress  =>
         "goodsRecords.downloadLinkText.fileInProgress"
-      case Some(downloadDataSummary) if downloadDataSummary.status == FileReady      =>
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileReadyUnseen =>
         "goodsRecords.downloadLinkText.fileReady"
-      case _                                                                         =>
+      case Some(downloadDataSummary) if downloadDataSummary.status == FileReadySeen   =>
+        "goodsRecords.downloadLinkText.fileReady"
+      case _                                                                          =>
```
liandra-giacomine: should this be the case when the option is empty?

besscerule: If the option is empty the have never requested a file so they are at the 'RequestFile' point in the journey

app/controllers/FileReadyController.scala
```diff
@@ -16,33 +16,58 @@
...
-      // TODO: Get file download link to pass in to view
-      // TODO: Get file created date and available until date
-      val fileSizeKilobytes = 1024
-      val fileDownloadLink  = "www.example.com"
-      val createdDate       = "19 July 2024"
-      val availableUntil    = "18 August 2024"
-      Ok(view(fileSizeKilobytes, fileDownloadLink, createdDate, availableUntil))
+      (for {
+        Some(downloadDataSummary) <- downloadDataConnector.getDownloadDataSummary(request.eori)
+        if (downloadDataSummary.status == FileReadyUnseen || downloadDataSummary.status == FileReadySeen) && downloadDataSummary.fileInfo.isDefined
```
ChrisCookOC: Could put this if logic in a pithily titled routine if you want. Might be more readable?

app/controllers/FileReadyController.scala
```diff
@@ -16,33 +16,58 @@
...
+        _                         <- downloadDataConnector.submitDownloadDataSummary(
+                                       DownloadDataSummary(request.eori, FileReadySeen, downloadDataSummary.fileInfo)
+                                     )
+        Some(downloadData)        <- downloadDataConnector.getDownloadData(request.eori)
+      } yield Ok(
+        view(
+          downloadDataSummary.fileInfo.get.fileSize,
```
ChrisCookOC: Can the fileInfo be missing?

Perhaps can move the .get to the yield so the error is handled? and then the .get is only in one place too

        Some(fileInfo) <- Future.successful(downloadDataSummary.fileInfo)

app/controllers/GoodsRecordsController.scala
```diff
@@ -102,25 +102,29 @@ class GoodsRecordsController @Inject() (
 
   def getDownloadLinkMessagesKey(opt: Option[DownloadDataSummary]): String =
     opt match {
-      case Some(downloadDataSummary) if downloadDataSummary.status == RequestFile    =>
+      case Some(downloadDataSummary) if downloadDataSummary.status == RequestFile     =>
```
ChrisCookOC: Don't need to do this, but perhaps this could be written slightly simpler like

opt.map(_.status) match {
case Some(FileInProgress) => 
case Some(FileReadyUnseen) => 
case Some(FileReadySeen) => 
case _ 
}

ChrisCookOC (approved):

PR #412, opened 2024-09-12
TGP-2298: create ukims cya

besscerule:


app/controllers/CyaCategorisationController.scala
```diff
@@ -70,7 +68,10 @@ class CyaCategorisationController @Inject() (
             }
             .getOrElse {
               dataCleansingService.deleteMongoData(request.userAnswers.id, CategorisationJourney)
-              logErrorsAndContinue("Failed to get categorisation details", continueUrl(recordId))
+              logErrorsAndContinue(
+                "Failed to get categorisation details",
+                routes.CategorisationPreparationController.startCategorisation(recordId)
```
ChrisCookOC, line 73: Doesn't particularly matter, just wondering if there is a reason to change all these

besscerule: Because there is weird behaviour with the continue url when it is a private val - it misses off part of the url

mycp98 (approved):
Looks really good to me!

PR #413, opened 2024-09-12
TGP-2354: Create NIPHL registered check your answers page

mycp98:


app/controllers/CyaMaintainProfileController.scala
```diff
@@ -74,6 +74,34 @@ class CyaMaintainProfileController @Inject() (
...
+    TraderProfile.validateHasNiphl(request.userAnswers) match {
+      case Right(_)     =>
+        val list = SummaryListViewModel(
+          rows = Seq(
+            HasNiphlSummary.rowUpdate(request.userAnswers)
+          ).flatten
+        )
+        Ok(view(list, routes.CyaMaintainProfileController.onSubmitNiphl))
+      case Left(errors) =>
+        logErrorsAndContinue(errorMessage, continueUrl, errors)
```
besscerule: the continue url doesn't work on the journey recovery page - check https://github.com/hmrc/trader-goods-profiles-frontend/pull/412 for a fix

app/controllers/HasNiphlController.scala
```diff
@@ -82,27 +83,30 @@ class HasNiphlController @Inject() (
...
+            traderProfileConnector.getTraderProfile(request.eori).flatMap { traderProfile =>
               for {
-                updatedAnswers <- Future.fromTry(request.userAnswers.set(HasNiphlUpdatePage, value))
-                _              <- sessionRepository.set(updatedAnswers)
-              } yield Redirect(navigator.nextPage(HasNiphlUpdatePage, NormalMode, updatedAnswers))
+                updatedAnswers                  <- Future.fromTry(request.userAnswers.set(HasNiphlUpdatePage, value))
+                updatedAnswersWithTraderProfile <-
+                  Future.fromTry(updatedAnswers.set(TraderProfileQuery, traderProfile))
+                _                               <- sessionRepository.set(updatedAnswersWithTraderProfile)
+              } yield Redirect(navigator.nextPage(HasNiphlUpdatePage, NormalMode, updatedAnswersWithTraderProfile))
```
besscerule: should this be mode instead of Normal Mode?

app/controllers/NiphlNumberController.scala
```diff
@@ -141,7 +141,7 @@ class NiphlNumberController @Inject() (
                         } yield Redirect(navigator.nextPage(NiphlNumberUpdatePage, NormalMode, answers))
                       case Left(errors) =>
                         val errorMessage = "Unable to update Trader profile."
-                        val continueUrl  = routes.HasNiphlController.onPageLoadUpdate
+                        val continueUrl  = routes.HasNiphlController.onPageLoadUpdate(NormalMode)
```
besscerule: this needed to be removed as a private variable and used within the code

app/models/TraderProfile.scala
```diff
@@ -104,4 +104,26 @@ object TraderProfile {
       case Left(errors) => Left(errors)
     }
 
+  def validateHasNiphl(
```
besscerule: what is the difference between buildNiphl and validate Niphl - i feel we should be using buildNiphl and not separating it out into has and number

app/models/ott/CategorisationInfo.scala
```diff
@@ -34,7 +34,7 @@ final case class CategorisationInfo(
   measurementUnit: Option[String],
   descendantCount: Int,
   longerCode: Boolean = false,
-  isTraderNiphlsAuthorised: Boolean = false,
```
besscerule, line 37: THANK YOU FOR CORRECTING THIS!!!!

app/navigation/Navigator.scala
```diff
@@ -223,6 +223,15 @@ class Navigator @Inject() (categorisationService: CategorisationService) {
       }
       .getOrElse(routes.ProfileController.onPageLoad())
 
+  private def navigateFromRemoveNiphlPage(answers: UserAnswers): Call =
+    answers
+      .get(RemoveNiphlPage)
+      .map {
+        case true  => routes.CyaMaintainProfileController.onPageLoadNiphl
+        case false => routes.ProfileController.onPageLoad()
```
besscerule, line 245: maybe leave to do to change this with the niphl number work?

app/navigation/Navigator.scala
```diff
@@ -439,6 +458,8 @@ class Navigator @Inject() (categorisationService: CategorisationService) {
     case HasNirmsUpdatePage                        => navigateFromHasNirmsUpdate
     case HasNiphlPage                              => navigateFromHasNiphlCheck
     case NiphlNumberPage                           => _ => routes.CyaCreateProfileController.onPageLoad
+    case RemoveNiphlPage                           => navigateFromRemoveNiphlPage
```
besscerule: i think this should be for the 'check' version of these functions (see other examples in this match)

mycp98: we use the same method for both normal and check mode, i can create a navigateFromRemoveNiphlPagecheck method but it would be the same code?

app/controllers/HasNiphlController.scala
mycp98: yep, i've changed it to mode now

app/navigation/Navigator.scala
besscerule: hmm good point - if mode doesn't matter why do we enter into the check version of this at all?

i guess if you check other examples there is always a check even if it is the same code? I'm a big fan of following the pattern that is already there - or changing it all to a new pattern if the old pattern is bad, it feels a little risky to have one pattern for one way but another pattern for another way

app/models/TraderProfile.scala
mycp98: We use validateHasNiphl because we are just validating the current data and we don't need to build a new TraderProfile which is what buildNiphl does, if that makes sense?

app/navigation/Navigator.scala
ChrisCookOC: I assume to a) Because the page we are on has a checkmode, but where we go doesn't depend on that.

Which kinda doesn't sound right to me tbh. Not sure we should be breaking out the journey if we were at CYA to go back to Profile Controller.

app/models/TraderProfile.scala
besscerule: it does - it's just build Niphl already does that - it only returns a built trader profile if it passes the tests - also we can use just one validation for both niphl number and has niphl

besscerule: what is validate niphl returning?

app/navigation/Navigator.scala
besscerule: i agree - we need to think about the pattern we are following and which one best suits - if we are not basing it on mode then these pages shouldn't be able to be in the CheckMode case match at all

besscerule: i feel like remove niphl shouldn't have a 'check' url

app/models/TraderProfile.scala
mycp98: you're right I've updated this now to use buildNiphl instead 😄

app/navigation/Navigator.scala
mycp98: RemoveNiphl doesnt have check mode so ive removed this now

app/navigation/Navigator.scala
```diff
@@ -223,6 +223,15 @@ class Navigator @Inject() (categorisationService: CategorisationService) {
       }
       .getOrElse(routes.ProfileController.onPageLoad())
 
+  private def navigateFromRemoveNiphlPage(answers: UserAnswers): Call =
+    answers
+      .get(RemoveNiphlPage)
+      .map {
+        case true  => routes.CyaMaintainProfileController.onPageLoadNiphl
+        case false => routes.ProfileController.onPageLoad() //todo: update this case when creating the niphl number cya
```
ChrisCookOC: niphl number cya has been merged already

ChrisCookOC (dismissed):

besscerule:
Bug found

To reproduce:
hasNiphls is no and you are on maintain profile page

click change hasniphls

go through journey to change hasNiphls to yes and niphls number to something and get to CYA page

click chnage hasNiphls on cya page

chnage hasNiphls to No

click continue

redirected to journey recovery page - should be redirect to CYA with just the No on hasNiphls

ChrisCookOC (approved):

PR #414, opened 2024-09-12
TGP-2344 Nirms number CYA

Ewan-James-Donovan:
PLEASE NO MERGING UNTIL UI TESTS ARE LOOKED AT

### DIFF
--- a/app/controllers/CyaMaintainProfileController.scala
+++ b/app/controllers/CyaMaintainProfileController.scala
@@ -48,19 +48,24 @@ class CyaMaintainProfileController @Inject() (
   private val errorMessage: String = "Unable to update Trader profile."
   private val continueUrl: Call    = routes.ProfileController.onPageLoad()
 
-  def onPageLoadNirmsNumber(): Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
-    request.userAnswers.get(NirmsNumberUpdatePage) match {
-      case Some(nirmsNumberAnswer) =>
-        val list = SummaryListViewModel(
-          rows = Seq(
-            Some(HasNirmsSummary.row(value = true, CheckMode)),
-            NirmsNumberSummary.row(Some(nirmsNumberAnswer))
-          ).flatten
-        )
-        Ok(view(list, routes.CyaMaintainProfileController.onSubmitNirmsNumber))
-      case None                    =>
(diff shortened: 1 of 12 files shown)


app/controllers/CyaMaintainProfileController.scala
```diff
@@ -0,0 +1,111 @@
...
+      traderProfileConnector.getTraderProfile(request.eori).flatMap { traderProfile =>
+        val updatedProfile = traderProfile.copy(nirmsNumber = request.userAnswers.get(NirmsNumberUpdatePage))
+        auditService.auditMaintainProfile(traderProfile, updatedProfile, request.affinityGroup)
+        for {
+          _ <- traderProfileConnector.submitTraderProfile(updatedProfile, request.eori)
+        } yield Redirect(navigator.nextPage(CyaMaintainProfilePage, NormalMode, request.userAnswers))
+      }
+  }
+
+  def onPageLoadNirms(): Action[AnyContent] = (identify andThen getData andThen requireData) { implicit request =>
```
Ewan-James-Donovan, line 82: This method and those below remain unchanged, I put the new methods at the top. I can reverse this the diff is confusing.

app/controllers/CyaMaintainProfileController.scala
```diff
@@ -47,6 +48,37 @@ class CyaMaintainProfileController @Inject() (
...
   private val continueUrl: Call    = routes.ProfileController.onPageLoad()
 
+  def onPageLoadNirmsNumber(): Action[AnyContent] = (identify andThen getData andThen requireData).async {
+    implicit request =>
+      traderProfileConnector.getTraderProfile(request.eori).flatMap { _ =>
+        request.userAnswers.get(NirmsNumberUpdatePage) match {
+          case Some(nirmsNumberAnswer) =>
+            val list = SummaryListViewModel(
+              rows = Seq(
+                Some(HasNirmsSummary.row(value = true, CheckMode)),
```
Ewan-James-Donovan, line 58: Hardcoding true is safe since it avoids url-hacking issues where a user could enter false, submit, and then url hack to cya page without confirming their hasnirms as false. Only the nirms number is altered on submit, if the user enters false for hasnirms and continues they go through the nirms removal path and the value is handled there separately. No worries about any bugs from me after manual testing.

Please give it a try manually if concerned

PR #418, opened 2024-09-16
Create NIPHL number check your answers page

liandra-giacomine:


app/models/TraderProfile.scala
```diff
@@ -104,4 +104,23 @@ object TraderProfile {
       case Left(errors) => Left(errors)
     }
 
+  def validateNiphlsUpdate(
```
besscerule, line 100: why aren't we using build niphls for this (the other function in this file)

app/controllers/CyaMaintainProfileController.scala
```diff
@@ -74,6 +74,35 @@ class CyaMaintainProfileController @Inject() (
...
+      case Left(errors) =>
+        logErrorsAndContinue(errorMessage, continueUrl, errors)
+    }
+  }
 
+  def onSubmitNiphls(): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
+    TraderProfile.validateNiphlsUpdate(request.userAnswers) match {
+      case Right(niphlNumber) =>
+        traderProfileConnector.getTraderProfile(request.eori).flatMap { traderProfile =>
+          val updatedProfile = traderProfile.copy(niphlNumber = niphlNumber)
```
besscerule: if we use build niphls we don't need to build the trader profile here because the function already does this

liandra-giacomine: It seems to end up complicating things rather than simplifying them

besscerule: Fair enough - could you also remove the build niphls and tests if it no longer used?

app/models/TraderProfile.scala
liandra-giacomine: We don't need to retrieve the profile for the page load

app/controllers/CyaMaintainProfileController.scala
liandra-giacomine: Good point, removed

app/controllers/RemoveNiphlController.scala
```diff
@@ -42,18 +41,22 @@ class RemoveNiphlController @Inject() (
   requireData: DataRequiredAction,
   profileAuth: ProfileAuthenticateAction,
   formProvider: RemoveNiphlFormProvider,
-  traderProfileConnector: TraderProfileConnector,
+  @unused traderProfileConnector: TraderProfileConnector,
```
ChrisCookOC: Can we just remove if it is unused now?

ChrisCookOC (dismissed):

Ewan-James-Donovan:
Would approve after merge conflicts fixed :)

app/controllers/RemoveNiphlController.scala
liandra-giacomine: Good shout, done

ChrisCookOC (approved):

PR #421, opened 2024-09-18
Fixed redirect for check your answer profile update

liandra-giacomine:


besscerule:
Bug found

To reproduce:
Niphls number is blank when click change from CYA page - it should contain the value shown on the CYA page

besscerule (approved):

PR #424, opened 2024-09-19
TGP-2414: Handle spaces in country of origin field (P1)

besscerule:


ChrisCookOC (changes requested):
Misunderstanding of the ticket I think! 

The answer can't have spaces in because its a 2 digit country code. 

This is about being able to enter spaces in the drop-down e.g. " Italy" or "Germany "

There is a comment on ticket with link to Slack where some people discuss implementing it for example

PR #438, opened 2024-09-23
TGP-2480: Update warning text placement on create profile cya page

mycp98:


Ewan-James-Donovan (approved):

besscerule:
Just a comment, the issue on git hub: https://github.com/hmrc/accessibility-audits-external/issues/8312 could imply that the warning text needs to change (see final comment). Is this the case? (if not just ignore :P)

ChrisCookOC:
I think it's fine as is @besscerule - my reading of that is just a short hand for our warning message (which does say that), and if there was a content change then we'd get a separate ticket anyway :)

ChrisCookOC (approved):

PR #467, opened 2024-09-30
TGP-2524: sup unit audit events

besscerule:


app/controllers/HasSupplementaryUnitController.scala
```diff
@@ -106,11 +109,21 @@ class HasSupplementaryUnitController @Inject() (
         .bindFromRequest()
         .fold(
           formWithErrors => Future.successful(BadRequest(view(formWithErrors, mode, recordId, onSubmitAction))),
-          value =>
+          value => {
```
besscerule: so on the others this is in the submit - so i put it here too - but i think i should be in the onPageload what do you think?

Ewan-James-Donovan: I say keep it here and open the discussion on  "Does it 'start' at the submit point or 'start' when the page loads?" - Kara may have an opinion

besscerule: we might need some tech debt to change it in other situations then

ChrisCookOC: needs to be onPageLoad. Has moved to onPageLoad in other scenarios on a earlier ticket this morning

app/factories/AuditEventFactory.scala
```diff
@@ -215,6 +215,29 @@ case class AuditEventFactory() {
...
+    recordId: String
+  )(implicit hc: HeaderCarrier): DataEvent = {
+    val auditDetails = Map(
+      "journey"              -> journey.toString,
+      "updateSection"        -> SupplementaryUnitUpdate.toString,
+      "recordId"             -> recordId,
+      "eori"                 -> supplementaryRequest.eori,
+      "affinityGroup"        -> affinityGroup.toString,
+      "addSupplementaryUnit" -> supplementaryRequest.hasSupplementaryUnit.getOrElse(false).toString,
+      "supplementaryUnit"    -> (if (supplementaryRequest.hasSupplementaryUnit.getOrElse(false)) {
```
ChrisCookOC: Should not set the field if supplementaryUnit is not set, rather than empty string it. 

"supplementaryUnit" -> supplementaryRequest.supplementaryUnit

should be fine because it shouldn't be set if hasSupplementaryUnit is not set anyway?

besscerule: we concatenate it with measurement unit so we need an if else

AmarkumarJagtap (approved):

PR #468, opened 2024-09-30
TGP-2197 - initial event show pages that could be shown to user

ChrisCookOC:


app/factories/AuditEventFactory.scala
```diff
@@ -93,7 +93,9 @@ case class AuditEventFactory() {
       writeOptional("recordId", recordId) ++
       writeOptional("commodityCode", commodity.map(_.commodityCode)) ++
       writeOptional("countryOfOrigin", commodity.map(_.countryOfOrigin)) ++
-      writeOptional("descendants", commodity.map(_.descendantCount.toString))
+      writeOptional("descendants", commodity.map(_.descendantCount.toString)) ++
+    // How many pages COULD be shown to the user
+      writeOptional("categoryAssessments", commodity.map(_.categoryAssessmentsThatNeedAnswers.size.toString))
```
AmarkumarJagtap, line 98: if there are no category assessments needed to answer, hope `categoryAssessmentsThatNeedAnswers` this won't return null otherwise it may throw null pointer exception for size

ChrisCookOC: Will be zero.

AmarkumarJagtap (approved):

PR #500, opened 2024-10-11
TGP-2660/fix warnings

Reece-Carruthers:
# Checklist for developers before reviewing
 - [x] I have merged in main
 - [x] I have run  `sbt prePR` to run all tests, format and check the code coverage
 - [x] I have written unit tests to cover any code that I wrote, where necessary
 - [x] I have ran the microservices with the latest SM2 config
 - [x] I have checked all the Acceptance Criteria pass
 - [x] I have run the pipeline journey tests
 - [x] I have run the local journey tests
 - [x] I have moved the ticket to Review in Jira
 - [x] I have left a message in tgp-prs on slack to alert other Devs that this ticket is ready for review

# Checklist for developers before merging
 - [x] The PR has been approved by 2 other developers
 - [x] I have merged in main
 - [x] I have run  `sbt preMerge' to run all tests, check formatting and code coverage
 - [x] I have ran the microservices with the latest SM2 config
 - [x] I have checked all the Acceptance Criteria pass
 - [x] I have run the pipeline journey tests
 - [x] I have run the local journey tests
 - [x] I have spoken to the QAs about any changes that may need to be made to the journey tests OR no changes to the journey tests are necessary

# Reminders for developers after merging
 1. Wait for pipeline to pass
 2. Move ticket to QA column in Jira
 3. Leave message in tgp-qa on slack to alert QAs that this ticket is ready for testing


app/connectors/DownloadDataConnector.scala
```diff
@@ -59,9 +61,10 @@ class DownloadDataConnector @Inject() (config: FrontendAppConfig, httpClient: Ht
         .map { response =>
           response.status match {
             case OK => Some(response.json.as[DownloadDataSummary])
+            case _  => None
           }
         }
-        .recover { case _: NotFoundException =>
+        .recover { case _: UpstreamErrorResponse =>
```
besscerule: we only want to do not found if there is a not found exception - else we want to let global error handle it

app/connectors/DownloadDataConnector.scala
```diff
@@ -75,20 +78,23 @@ class DownloadDataConnector @Inject() (config: FrontendAppConfig, httpClient: Ht
       .map { response =>
         response.status match {
           case OK => Some(response.json.as[DownloadData])
+          case _  => None
         }
       }
-      .recover { case _: NotFoundException =>
-        None
+      .recover {
```
besscerule, line 85: ahh can you do this on the one above

app/controllers/RemoveNiphlController.scala
```diff
@@ -71,6 +71,7 @@ class RemoveNiphlController @Inject() (
                   sessionRepository.set(answers).map { _ =>
                     Redirect(navigator.nextPage(RemoveNiphlPage, NormalMode, answers))
                   }
+                case _                => Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url))
```
besscerule, line 74: is this a JR choice according to what @oliverslade said earlier or should we let the global error handler handle it?

app/controllers/RemoveNirmsController.scala
```diff
@@ -71,6 +71,7 @@ class RemoveNirmsController @Inject() (
                   sessionRepository.set(answers).map { _ =>
                     Redirect(navigator.nextPage(RemoveNirmsPage, NormalMode, answers))
                   }
+                case _                => Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url))
```
besscerule, line 74: see above

besscerule:
Just a few comments 

In general though this PR makes me ridiculously happy - looking forward to no warnings! 

Can you also add it as a checkpoint on the Pre PR checklist

app/controllers/RemoveNiphlController.scala
Reece-Carruthers: We still need to have something locally to handle the incomplete match otherwise Scala will complain that there is missing cases

besscerule (approved):

app/controllers/RemoveNiphlController.scala
besscerule: Hmmm can we ask oil about it? It might be a case that we can just hide these warnings

pauloncast (approved):

app/controllers/RemoveNiphlController.scala
oliverslade: @besscerule @Reece-Carruthers can you briefly explain what the scenario is where we are using JourneyRecovery?

Reece-Carruthers: @oliverslade basically to solve some warning issues with incomplete matches I decided to add a case which will catch any failures and redirect the user to the journey recovery

The scenario in question is what would we do in the event of user answers failing to set an answer 

```
if (value) {
              request.userAnswers.set(RemoveNiphlPage, value) match {
                case Success(answers) =>
                  sessionRepository.set(answers).map { _ =>
                    Redirect(navigator.nextPage(RemoveNiphlPage, NormalMode, answers))
                  }
                case _                => Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad().url))
              }

besscerule: in this case when a user submits and answer to this page (which is whether you want to remove a niphl number (which is a filed that is associated with the user's profile) we set the userAnswers (standard pattern from the scaffold) to true or false for that field. If we set it successfully (which is guess is the most likely option) everything is good - if we don't we now go to the journey recovery page instead of the global error handler. Whether or not we set it successfully depends on if it is not a JSObject I think. So I'm not sure if that counts as a place to use Journey Recovery. Because if the user keeps trying this again it will break because they can only do true or false so if it errors it means we have a bug not that the user is wrong. I guess what I am saying is this doesn't feel like a path the user should ever even be on because if they get here something is broken.

oliverslade: Taking a look, theres no hard and fast rule on this. Quite often its used as a fall back for many/any/some/all errors. And some teams use it differently, sometimes you know whats gone wrong and its recoverable, sometimes you dont. Something you can do is route the user back to the start/home page/tasklist or whatever and hope they resolve whatever has gone wrong themselves. OR clear userAnswers at that point (telling the user/giving them the option) for something that might be unresolvable.
I think what you have done here is fine. The question seems like it is a yes/no form so if the value they set is something different, it must mean that something has gone wrong or theyve manipulated the request.
To add to what I mentioned yesterday. If we get to the end of the journey and validated the full user answers into a payload before we send it, and notice the niphl number boolean was not set, we know that the validation will fail, in which case the submission wont happen and we know why. I think in this case its better to go to journey recovery here so the user can potentially resolve the issue. Where as the global error handler error page is a dead end of the user.

